PR #30, opened 2017-04-14
Fixes #29 - Provide context menu for discover mode

### LINKED ISSUE #29: Provide context menu for discover mode
It would be nice to have an overview of the possible keybinds+commands through a magit-like popup.

Possible options are:
* https://github.com/steckerhalter/discover-my-major
* https://github.com/mickeynp/discover.el

I'll provide a PR for both.

fapdash:
* provides a `discover.el` context menu
* moves lambdas from keymap to functions so `discover-my-major` can pick
  them up

### DIFF
--- a/README.org
+++ b/README.org
@@ -19,6 +19,14 @@ Then, require it and go.
 I'll make mastdon.el available on MELPA when I feel like it's reached a
 stable state.
 
+Mastodon mode provides a overview of it's shortcuts through [[https://github.com/mickeynp/discover.el][Discover mode]].
+To activate Discover mode locally for Mastodon mode buffers add this to your emacs configuration:
+#+BEGIN_SRC emacs-lisp
+    (add-hook 'mastodon-mode-hook 'discover-mode-turn-on)
+#+END_SRC
+
+The context menu is then activated by pressing '?'.
+
 ** Usage
 
 *** Instance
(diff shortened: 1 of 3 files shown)


jdenen (changes requested):
Hmm. I fixed my issue with `use-package`.

```emacs-lisp
(use-package mastodon
  :init (discover-mode t)
  :load-path "~/path/to/mastodon.el/lisp")
```

Using `(when (featurep 'discover) ... )` requires `discover` to be global. I'd rather not force users to use `discover` globally to get this feature. So I propose:

1. Wrapping the entire `(when (featurep 'discover) ... )` call in a function. 
2. Adding that function to the `mastodon-mode-hook` with the APPEND flag non-nil.

This _should_ work when `discover` is global and it _should_ let non-global users opt into the `discover` feature with:

```emacs-lisp
(add-hook 'mastodon-mode-hook 'discover-mode-turn-on)
```

fapdash:
Mh, now the context menu is only available after running `mastodon` twice.
I've run out of time to debug this but I'll look at it again next week.

@jdenen Do you have an idea? I already tried use `add-advice` instead but it didn't seem to fix the issue.

jdenen:
I'll take another look tonight. If I figure it out, I'll cherry pick your commit and add to it. 

I want to get this in before I add more commands to the mode map.

PR #64, opened 2017-04-24
Partial fix #41

### LINKED ISSUE #41: Display media in timelines
They just display as links right now.

alexjgriffith:
I have added inline media links and content warnings.

1. Media Links
    - currently the media links are displayed below the content of the toot. To view the link you can use `M-x ffap RET`
    - I have a prototype for inline images, but currently it is very buggy and slows down page loading considerably
    - Note: When the server can't find the requested image it returns a default string. This default string should be detected and return a local image that is distributed with the package

 2. Content warnings
    - The content warnings now appears above their toots. 
    - Currently there is no folding, however it could be implemented using something like hideshow

### DIFF
--- a/lisp/mastodon-tl.el
+++ b/lisp/mastodon-tl.el
@@ -140,6 +140,44 @@ Return value from boosted content if available."
      'toot-id id
      'toot-json toot)))
 
+
+(defun mastodon-tl--set-face (string face render)
+  "Set the face of a string. If `render` is not 'nil
+also render the html"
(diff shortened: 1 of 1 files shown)


jdenen:
Can I close this PR in favor of #67? It looks like it's included.

alexjgriffith:
Yes, #67 will supersede this branch entirely.

jdenen:
Sounds good.

PR #66, opened 2017-04-24
Fixed #65 pointer placement when getting more toots

### LINKED ISSUE #65: mastodon-tl--more jumps to final toot
When requesting more toots `mastodon-tl--more` results in the point being set to the end of the buffer rather than at the first new toot.

alexjgriffith:
`Point` will now fall on the newest loaded toot.

If the previous behavior, i.e., `Point` ending at `point-max`, is preferred then we should work this into an option.

### DIFF
--- a/lisp/mastodon-tl.el
+++ b/lisp/mastodon-tl.el
@@ -217,14 +217,17 @@ Return value from boosted content if available."
 (defun mastodon-tl--more ()
   "Append older toots to timeline."
   (interactive)
-  (let* ((tl (mastodon-tl--timeline-name))
+  (let* ((point-before-update (point))
+	 (tl (mastodon-tl--timeline-name))
          (id (mastodon-tl--oldest-id))
          (json (mastodon-tl--more-json tl id)))
     (when json
       (with-current-buffer (current-buffer)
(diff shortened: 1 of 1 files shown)


lisp/mastodon-tl.el
```diff
@@ -217,14 +217,17 @@ Return value from boosted content if available."
 (defun mastodon-tl--more ()
   "Append older toots to timeline."
   (interactive)
-  (let* ((tl (mastodon-tl--timeline-name))
+  (let* ((point-before-update (point))
+	 (tl (mastodon-tl--timeline-name))
```
jdenen: Can you fix the spacing on this line? I know it's a little thing but might as well clean it up.

jdenen (changes requested):

alexjgriffith:
I think the indentation is off due to the diff format.

On my machine and in the repo it looks like this
screenshot 2017-04-24 12 29 52

screenshot 2017-04-24 12 31 46

jdenen (approved):
Weird. When I apply the patch, the indentation looks wrong on my machine. But it's not a big deal. The functionality looks great. Thanks!

PR #72, opened 2017-04-24
Use simpler `json-read`

hdurer:
Don't use the pattern `(json-read-from-string (buffer-substring-no-properties (point) (point-max)))` where `(json-read)` will do the same thing but more efficiently.

`json-read-from-string` inserts the argument into a temp buffer and then calls `json-read` on that, i.e. with the old pattern we only consed a lot of temporary memory with no added benefit.

### DIFF
--- a/lisp/mastodon-auth.el
+++ b/lisp/mastodon-auth.el
@@ -58,9 +58,8 @@
     (re-search-forward "^$" nil 'move)
     (let ((json-object-type 'plist)
           (json-key-type 'keyword)
-          (json-array-type 'vector)
-          (json-string (buffer-substring-no-properties (point) (point-max))))
-      (json-read-from-string json-string))))
+          (json-array-type 'vector))
+      (json-read))))
 
 (defun mastodon-auth--access-token ()
   "Return `mastodon-auth--token'.
--- a/lisp/mastodon-client.el
+++ b/lisp/mastodon-client.el
@@ -56,9 +56,8 @@
     (re-search-forward "^$" nil 'move)
     (let ((json-object-type 'plist)
           (json-key-type 'keyword)
-          (json-array-type 'vector)
-          (json-string (buffer-substring-no-properties (point) (point-max))))
-      (json-read-from-string json-string))))
+          (json-array-type 'vector))
+      (json-read))))
 
 (defun mastodon-client--token-file ()
   "Return `mastodon-token-file'."
--- a/lisp/mastodon-http.el
+++ b/lisp/mastodon-http.el
@@ -104,8 +104,7 @@ Pass response buffer to CALLBACK function."
          (with-current-buffer (mastodon-http--get url)
            (goto-char (point-min))
            (re-search-forward "^$" nil 'move)
-           (let ((json-string (buffer-substring-no-properties (point) (point-max))))
-             (json-read-from-string json-string)))))
+           (json-read))))
     json-vector))
 
 (provide 'mastodon-http)


jdenen:
I haven't forgotten about this PR. When I cherry-picked it, my client stopped rendering non-ascii characters. But I'm not sure if this is a problem with the PR or my setup at that time. Looking at the code, I don't see why it would be a problem with the PR.

So, I've got a few more things to do before I take a hard look into why this is happening to me. I should get around to it in the next day or so.

Thanks for the contribution! It looks like a nice improvement. 👍

jdenen:
After looking into it a bit more, this seems to be introducing problems with utf-8 characters:

screen shot 2017-04-27 at 11 38 12 pm

The only emoji that is displaying is one written the GitHub way (`:sweat-smile:`). The weirdly rendered characters on that same line should also be displayed as 😅.

jdenen (changes requested):
I should have left that last comment in a review. This comment can be ignored.

hdurer:
Ack.  I'll have a look.

There is obviously some weirdness going on w.r.t. text encoding and defaults.

Is Mastodon guaranteeing us that they send UTF-8?  I was already confused by the `decode-coding-string string` calls.  If it is utf-8 maybe we should tell emacs early on that the buffer is in that encoding and stop doing the decodings...

I'll do some investigating.

alexjgriffith:
making the http call:

``` elsip
(switch-to-buffer-other-window(mastodon-http--get (mastodon-http--api "timelines/public")) )
```

We defiantly have to manually decode the output of url-request to utf-8 even though it includes the charset in the header

```
Content-Type: application/json; charset=utf-8
```

the output of `M-x describe-char` in the http buffer

```
             position: 799 of 33195 (2%), column: 5
            character: " (displayed as ") (codepoint 34, #o42, #x22)
    preferred charset: ascii (ASCII (ISO646 IRV))
code point in charset: 0x22
               script: latin
               syntax: " 	which means: string
             category: .:Base, a:ASCII, l:Latin, r:Roman
             to input: type "C-x 8 RET 22" or "C-x 8 RET QUOTATION MARK"
          buffer code: #x22
            file code: #x22
              display: by this font (glyph code)
    uniscribe:-outline-Courier New-normal-normal-normal-mono-13-*-*-*-c-*-iso8859-1 (#x05)

Character code properties: customize what to show
  name: QUOTATION MARK
  general-category: Po (Punctuation, Other)
  decomposition: (34) ('"')
```

alexjgriffith:
From the decoded buffer

```
             position: 1085 of 30053 (4%), column: 291
            character: ❎ (displayed as ❎) (codepoint 10062, #o23516, #x274e)
    preferred charset: unicode (Unicode (ISO10646))
code point in charset: 0x274E
               script: symbol
               syntax: w 	which means: word
             category: .:Base
             to input: type "C-x 8 RET 274e" or "C-x 8 RET NEGATIVE SQUARED CROSS MARK"
          buffer code: #xE2 #x9D #x8E
            file code: #xE2 #x9D #x8E (encoded by coding system utf-8-dos)
              display: no font available

Character code properties: customize what to show
  name: NEGATIVE SQUARED CROSS MARK
  general-category: So (Symbol, Other)
  decomposition: (10062) ('❎')

There are text properties here:
  face                 font-lock-string-face
  fontified            t

[back]

```

alexjgriffith:
## Bug Hunting
I can recreate the malformed utf-8 by running `recode-region` on the buffer before passing it to the original function.

## Functions
Altered `mastodon-inspect--dump-json-in-buffer` to enable `emojify-mode`

``` elisp
(defun mastodon-inspect--dump-json-in-buffer (name json)
  "Buffer NAME is opened and JSON in printed into it."
  (switch-to-buffer-other-window name)
  (progn (setf print-level nil
            print-length nil)
         (insert (pp json t))
         (goto-char 1)
         (emacs-lisp-mode)
	 (emojify-mode t)
         (message "success")))

```

The original function 
``` elisp
(defun mastodon-http--get-json-before (url)
   "Make GET request to URL. Return JSON response vector."
   (let ((json-vector
           (with-current-buffer (mastodon-http--get url)
             (goto-char (point-min))
             (re-search-forward "^$" nil 'move)
            (let ((json-string (buffer-substring-no-properties (point) (point-max))))
              (json-read-from-string json-string)))))
json-vector))
```

Removing the redundant buffer dump
``` elsip
 (defun mastodon-http--get-json-after (url)
   "Make GET request to URL. Return JSON response vector."
   (let ((json-vector
           (with-current-buffer (mastodon-http--get url)
             (goto-char (point-min))
             (re-search-forward "^$" nil 'move)
            (json-read))))
     json-vector))
```

Recreating the error of `mastodon--http-get-json-after by forcing the coding to utf-8

```elisp
(defun mastodon-http--get-json-with-forced-encoding (url)
  "Make GET request to URL. Return JSON response vector."
  (let ((json-vector
	 (with-current-buffer (mastodon-http--get url)
	   (recode-region 1 (point-max) 'utf-8 'raw-text)
	   (goto-char (point-min))
	   (re-search-forward "^$" nil 'move)	   
            (json-read))))
     json-vector))
```

## Checking the outputs

``` elisp
(defvar json-read-debug-api-call (mastodon-http--api "statuses/4610088"))

;; the raw http source
(switch-to-buffer-other-window
 (mastodon-http--get json-read-debug-api-call))

;; check the orig
(mastodon-inspect--dump-json-in-buffer
 "*test-before*"
 (mastodon-http--get-json-before json-read-debug-api-call))

;; check the post
(mastodon-inspect--dump-json-in-buffer
 "*test-after*"
 (mastodon-http--get-json-after json-read-debug-api-call))

;; recreate the bug in post by forcing recoding to utf-8
(mastodon-inspect--dump-json-in-buffer
 "*test-solution*"
 (mastodon-http--get-json-with-forced-encoding json-read-debug-api-call)) 
```

we get a similar issue if we run the following on the http buffer
``` elisp
(insert (decode-coding-region 1 (point-max) 'utf-8 t))
```

It looks like it does not decode groups

alexjgriffith:
temp-buffer creates a buffer with raw-text as its coding scheme, the url-receive buffer however is in the default coding-scheme for your distro (mine is iso-latin-1). I've included a sample of outputs and their conversions below. Unfortunatly github does not render "raw-text" to recreate this example take the display name from statuses/4610088. 

``` elisp

(defvar json-read-debug-api-call (mastodon-http--api "statuses/4610088"))

 (cdr
  (assoc ' display_name
	   (cdr (assoc 'account
		       (mastodon-http--get-json-before
			json-read-debug-api-call)))))
```

Raw output
``` elisp
;; "                "
```

iso-latin-1 output
``` elisp
;; ç´æ¨ããªð¦
```

utf-8 output
``` elisp
;; 直木ナオ🦊
```

output from the new version decoded as iso-latin-1 evaluates as itself
``` elisp
(insert(decode-coding-string "ç´æ¨ããªð¦" 'iso-latin-1))
;; ç´æ¨ããªð¦
```

output from the old version decoded as utx-8 evaluates
to the expected output
``` elisp
(insert(decode-coding-string "                "'utf-8))
;; 直木ナオ🦊
```

Recoding the region of the origional function as iso-latin-1 results
in the same output as the new function
``` elisp
(insert(with-temp-buffer (insert "                ") (recode-region 1 (point-max) 'iso-latin-1 'raw-text) (buffer-string)))
;; ç´æ¨ããªð¦
```

Recoding the region of the origional function as utx-8 results
in the expected output
``` elisp
 (insert(with-temp-buffer (insert "                ") (recode-region 1 (point-max) 'utf-8 'raw-text) (buffer-string)))
;; 直木ナオ🦊
```

Recoding the output of the new function as utf-8 from iso-latin1 results
in the expected output
``` elisp
(insert(with-temp-buffer (insert "ç´æ¨ããªð¦") (recode-region 1 (point-max) 'utf-8 'iso-latin-1) (buffer-string)))
;; 直木ナオ🦊
```

hdurer:
After some more investigation by various people in the matrix channel the conclusion was the we do need to somehow explicit decode the data and that extracting the string and using json-read-from-string was the best way forward.

Ergo we should drop this pull request.

PR #75, opened 2017-04-25
Fix #70 and added mastodon-toot--unfavorite

### LINKED ISSUE #70: Favourite/Boost moves point back one toot
If I Favourite/Boost the third toot on a timeline, my cursor will be moved to the second toot on the timeline. It should stay on the third.

alexjgriffith:
Two bugs were fixed
1. Point position
     - the point position now consistently appears at the toot being favorited
2. Position of (F) and (B)
     - There was a bug that would put the mark at the end of the toot byline rather than at the front. It now consistently adds the (F) and (B) to the front of the byline

For testing purposes I also added an unfavorite function and created a callback function when marks are to be removed from the byline.


jdenen:
I'm branching off this PR, and I'll push up a branch that will subsume it. I like the #70 fix but I want to change the approach to favourite/unfavourite. 

If I figure out how to, I'll configure it so you can push to the branch too.

jdenen:
Subsumed by #85

Issue #20, opened 2017-04-12
Show profile in major mode


jdenen:
I want to be able to visit a user profile like a timeline in the major mode.

alexjgriffith:
I have an initial version of user profile streams working.

The function `mastodon-user--get-next-author` opens a new buffer named `*mastodon-$user*`. Once the buffer is generated it inserts the users static image directly along with their note (description). 

## Layout
```
$avatar_static
$modified_byline ;; still working on
$note

$content
 | $user (@$acct)
  ------------

$content
 | $user (@$acct)
  ------------
```

## Limitations
Currently it works as well as the notifications screen. I.e. it has full functionality apart from loading new toots. For this a generalized version of `mastodon-tl--update` will have to be written. I have included this as an example in the matodon-notifications.el module.

## Example
Please note that I have not added the modified by-line yet
![screenshot 2017-04-25 01 25 36](https://cloud.githubusercontent.com/assets/6475397/25370061/1b30511e-2956-11e7-97df-c553b4e18d12.png)

PR #81, opened 2017-04-25
#20 Add profile interaction

alexjgriffith:
This is the first portion of the profile interaction.

I have bound `U` to `mastodon-profile--get-next-author`, which opens a new buffer containing the authors image, their name and description, and a list of their toots. When you open this buffer Point starts on the first toot.

Currently `mastodon-profile` and `mastodon-notifications` rely on `mastodon-tl` for their key functionality. While this works well for rendering it is not sufficient for the get more toots functionality.

The next step for the profile-module is to open a buffer of their followers and a buffer of those they follow. 

In addition it would be nice to have an indication of whether or not you are following a person. This could be incorporated into the by-line in the profile buffer.

### DIFF
--- a/Cask
+++ b/Cask
@@ -1,7 +1,7 @@
 (source gnu)
 (source melpa)
 
-(package "mastodon" "0.5.5" "Emacs client for Mastodon")
+(package "mastodon" "0.6.0" "Emacs client for Mastodon")
 (files "lisp/*.el")
 
 (development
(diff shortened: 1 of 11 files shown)


alexjgriffith:
It looks like the latest release does not play well with the format of user statuses on a timeline. 
```
Symbol's function definition is void: mastodon-tl--set-face
```
 Will push a fix tonight.

*EDIT:* Turns out it was because the master branch on my work PC was pre 0.5 so `mastodon-tl--set-face` had not been defined

## Example of mastodon-profile.el in action
profile-buffer

Issue #74, opened 2017-04-24
Auto fill Tag search with the string bellow point


alexjgriffith:
Add `word-at-point` to the `mastodon-tl--get-tag-timeline` function. This will set the word under point as default in case no tag is entered interactively.

```
M-x mastodon-tl--get-tag-timeline RET Tag($word): RET
```

jdenen:
Merged to master.

PR #82, opened 2017-04-25
en #74 added autofill for tag timeline

alexjgriffith:
Pressing `T` when in `mastodon-mode` will now include the word under Point as the default option.

### DIFF
--- a/lisp/mastodon-tl.el
+++ b/lisp/mastodon-tl.el
@@ -54,7 +54,10 @@
 (defun mastodon-tl--get-tag-timeline ()
   "Prompts for tag and opens its timeline."
   (interactive)
-  (let ((tag (read-string "Tag: ")))
+  (let* ((word (word-at-point))
+	 (input (read-string (format "Tag(%s): " word)))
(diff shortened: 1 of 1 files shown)


lisp/mastodon-tl.el
```diff
@@ -54,7 +54,10 @@
 (defun mastodon-tl--get-tag-timeline ()
   "Prompts for tag and opens its timeline."
   (interactive)
-  (let ((tag (read-string "Tag: ")))
+  (let* ((word (word-at-point))
```
jdenen: Currently, if you're `(word-at-point)` is nil, then the prompt looks like `Tag(nil): `.

It can be fixed with a small tweak:

```elisp
(word (or (word-at-point) ""))
```

This will look like `Tag(): `.

I'd like to optionally display the parens, but I think it can be fixed in later commits.

jdenen (changes requested):

jdenen (approved):

jdenen:
Ah, I can't believe I missed this one before the 0.6.1 release.

Issue #91, opened 2017-04-26
Support favorites list


clarkenciel:
Support actions against the [`GET /api/v1/favourites`](https://github.com/tootsuite/documentation/blob/master/Using-the-API/API.md#favourites) endpoint would be nifty.

jdenen:
This feels like a pretty easy thing to do. It seems like just another timeline.

Added it to the 1.0.0 milestone.

alexjgriffith:
Implementing this feature requires us to capture the `max_id` and the `since_id` in the response header, i.e. we will need to rework how we handle the response from `url-retrieve-synchronously`. 

Perhaps at the same time we can make the call asynchronous?

[From Mastodon API Documentation](https://github.com/tootsuite/documentation/blob/master/Using-the-API/API.md#favourites)

> Note: max_id and since_id for next and previous pages are provided in the Link header. It is not possible to use the id of the returned objects to construct your own URLs, because the results are sorted by an internal key.

Edit: See #197

Issue #84, opened 2017-04-25
Suggestion for get-more generalization


alexjgriffith:
## Suggestion
Could we generalize some functions in mastodon-tl.el including `mastodon-tl--more-json` and ` mastodon-tl--goto-next-toot`? Currently they only work with `/timeline` endpoints, however `/notifications` and `/accounts` can be processed to appear functionally equivalent. Keeping the buffers functionally equivalent can help make the interaction with each type of buffer consistent.

## Generalizing `mastodon-tl--more-json`
I propose pulling out the URL generation and call into a more general function in mastodon-http.

### Example
In mastodon-http.el we could add `mastodon-http--get-json-with-args` or an equivelent function,
``` elisp
(defun  mastodon-http--get-json-with-args (endpoint args)
  "Return JSON for ENDPOINT with ARGS, where ARGS is a plist."
  (let* ((formated-args (mapconcat
			 (lambda(x)
			   (concat (format "%s" (car x))
				   "="
				   (format "%s" (cdr x))))
			 args "&"))
	 (url (mastodon-http--api (concat endpoint 
					 "?"
					 formated-args))))
    (message url)
    (mastodon-http--get-json url)))

``` 

in mastodon-tl.el `mastodon-tl--more-json` could be rewriten as:
``` elisp
(defun mastodon-tl--more-json(timeline id)
  (mastodon-http--get-json-with-args (concat "timelines/" timeline)
				     `(("max_id".,id))))
```
and  `mastodon-tl--update-json`  as:
``` elisp
(defun mastodon-tl--more-json(timeline id)
  (mastodon-http--get-json-with-args (concat "timelines/" timeline)
				     `(("since_id".,id))))
```
This will let us add versions of more-json for `\accounts` and `\notifications` endpoints.

## Generalizing `mastodon-tl--goto-next-toot`
The best place for generalization is probably in the `mastodon-tl--goto-next-toot` function. This is necessary to call the appropriate more-json function and to provide it with the appropriate argument for the `text-properties` that are relevant in each buffer. 

Currently requesting a next toot `j` calls `mastodon-tl--goto-next-toot` directly as follows:
``` elisp
(defun mastodon-tl--goto-next-toot ()
  "Jump to next toot header."
  (interactive)
  (mastodon-tl--goto-toot-pos 'next-single-property-change
                              'mastodon-tl--more))
```

[...]

PR #92, opened 2017-04-26
adding favorites timeline functionality

clarkenciel:
attempt to address: https://github.com/jdenen/mastodon.el/issues/91

### DIFF
--- a/lisp/mastodon-tl.el
+++ b/lisp/mastodon-tl.el
@@ -58,6 +58,11 @@
   (let ((tag (read-string "Tag: ")))
     (mastodon-tl--get (concat "tag/" tag))))
 
+(defun mastodon-tl--get-favorites-timeline ()
+  "Opens 'timeline' of favorites."
+  (interactive)
+  (mastodon-tl--get-favorites))
+
 (defun mastodon-tl--goto-toot-pos (find-pos refresh &optional pos)
   "Search for toot with FIND-POS.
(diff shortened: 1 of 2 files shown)


clarkenciel:
hm okay this was a bit premature: i need to work on the "load-more functionality" for this since the defaults in the timeline module won't work out-of-the-box for this. @jdenen i'd rather not duplicate most of that module. how do you feel about adding adding a "fetch-more" callback to some of the timeline functions? I'd rather not have to duplicate pretty much the entire timeline module just because the endpoint for favorites is slightly different.

alexjgriffith:
@clarkenciel I had the same issues with notifications, user profiles, and now async streams. @jdenen mentioned his plan going forward in #84. We need to get those macros done in mastodon-tl.el before we get full functionality.

jdenen:
Yeah #84 is the approach we're going to take to extract all the commonality. Let's let this PR stand for a bit and refactor it once the #84 approach is implemented in develop. 

In the meantime, can you edit this PR to change all the isntances of "favorite" to "favourite"? 

While I'm American and do appreciate the _right_ way to spell it (😝), I want this to be as close to the vanilla Mastodon experience as possible. Eventually, we could add customization around a lot of these strings but that's gold plating that I'd rather put off for awhile.

clarkenciel:
@jdenen @alexjgriffith sounds like a plan y'all!

PR #96, opened 2017-04-27
#87 added support for inspecting toot under Point

### LINKED ISSUE #87: Tools for inspecting / debugging toots
## Suggestion
Here are some easy to use tools for seeing the properties of a toot. I've been using `mastodon-debug--inspect-toot` for inspecting / debugging toots. It pretty prints the results and dumps them into a new buffer. The focus shifts to the new buffer which is opened in another window.

I am not sure if this functionality has been implemented already, but I have not come across it. I am not sure if it should be included in the project itself. The functions can work stand alone when dropped in your init file.

``` elisp
(defun mastodon-debug--dump-json-in-buffer (name json)
  "Buffer NAME is opened and JSON in printed into it."
  (switch-to-buffer-other-window name)
  (progn (setf print-level nil
            print-length nil)
         (insert (pp json t))
         (goto-char 1)
         (emacs-lisp-mode)
         (message "success")))

(defun mastodon-debug--inspect-toot ()
  "Find next toot and dump its meta data into new buffer."
  (interactive)
  (mastodon-debug--dump-json-in-buffer
   (concat "*mastodon-debug-toot-"
           (int-to-string (mastodon-tl--property 'toot-id))
           "*")
  (mastodon-tl--property 'toot-json)))
```

I recommend binding `D` or `I` to  `mastodon-debug--inspect-toot` or its equivalent.

Currently that can be done by adding the following to your init file:
``` elisp
(define-key mastodon-mode-map (kbd "D") #'mastodon-debug--inspect-toot)
```

**NOTE:** Is this an appropriate place to make suggestions, or is their another place they should be submitted?

## Sample Screenshot
screenshot 2017-04-26 02 22 46

**EDIT** added screen shot.

alexjgriffith:
To close #87

This change adds a new module, mastodon-inspect. Currently it only supports inspecting toots under Point, but it can be expended for more debugging and inspection purposes.

The new interactive command that is provided is `mastodon-inspect--toot`. It will display the lisp contents of the current toot in a new buffer and then display that buffer.

You can add a binding in your init file like this.
``` elisp
(define-key mastodon-mode-map (kbd "D") #'mastodon-inspect--toot)
```

### DIFF
--- a/lisp/mastodon-inspect.el
+++ b/lisp/mastodon-inspect.el
@@ -1,0 +1,59 @@
+;;; mastodon-inspect.el --- Client for Mastodon
+
+;; Copyright (C) 2017 Johnson Denen
+;; Author: Johnson Denen <johnson.denen@gmail.com>
+;; Version: 0.6.0
+;; Package-Requires: ((emacs "24.4"))
+;; Homepage: https://github.com/jdenen/mastodon.el
+
+;; This file is not part of GNU Emacs.
+
+;; This file is part of mastodon.el.
+
+;; mastodon.el is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; mastodon.el is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


jdenen (approved):
You'll need to require in your init file too:

```elisp
(require 'mastodon-inspect)
(define-key mastodon-mode-map (kbd "D") #'mastodon-inspect--toot)
```

But that's as much documentation as this feature is going to get. It looks great and will be super helpful.

Issue #14, opened 2017-04-12
Documentation in mastodon-toot mode buffer


jdenen:
When you open a toot-mode buffer, there should be read-only documentation:

1. `C-c C-c` sends toot
2. `C-c C-k` cancels toot

jdenen:
Included in v0.6.1

Issue #15, opened 2017-04-12
Character counter in toot-mode buffer


jdenen:
Buffers for `toot-mode` should count down characters from 500 as they are entered.

It would be nice if the countdown lived alongside the documentation from #14.

clarkenciel:
some instance have different character counts (cybre.space has 512, witches.town has 666) is there even a way to handle this?

jdenen:
I didn't know that. 

Perhaps the answer then is to count up from 0 and let users decide how many is too many.

clarkenciel:
okay so the server will return a 422 if you submit a post that is too long and the body will say:
`{"error":"Validation failed: Text character limit of 512 exceeded"}`. So going with 500 will be fine if we assume that instances don't set limits to be lower. One thing we could think about is using the .social defaults and, on encountering an error like this (instances like oulipo.social or dolphin.town have other, more exotic validations) present it to the user, and where applicable add appropriate client-side validations; perhaps storing it for later.

For text length that would be easy, for the weirder instances, i dunno that's not our problem probably?

jdenen:
Yeah, not our problem. There's no way to sanely support the outlandish/hilarious rules each instance could implement. Best to just support the vanilla Mastodon experience while not blocking the ability for other experiences to use mastodon.el.

clarkenciel:
I'm working on this and i think i'm going to make it a customizable variable. We'll default to 500, but users can update it easily if they want to reflect their instance.

clarkenciel:
@jdenen @alexjgriffith okay since this is being referenced elsewhere, here's my suggestion for how this looks:

```
|=========================|
  doc text....
|=========================|

toot text [9 / 500]
```

update that counter next to the text as it's written. change the face of the counter to `'warning` if the current-count exceeds the `max-count`

jdenen:
The counter is going to move as the text is written? Because that sounds awesome.

jdenen:
What I mean by move, is the actual location in the buffer. So

`toot text [9 / 500]` becomes `toot text continued [18/ 500]` becomes:

```
toot text continued

to a new

line [33 / 500]
```

?

clarkenciel:
exactly the idea @jdenen !

jdenen:
Yeah I like that.

alexjgriffith:
Big fan of the inline counter! I now i get the impact on #14.

 I was wondering how we were going to make it consistent in the buffer mode line. I think this approach is much better!

[...]

PR #97, opened 2017-04-27
#14 adding keybinding documentation to the new-toot buffer

clarkenciel:
attempt at fixing https://github.com/jdenen/mastodon.el/issues/14

### DIFF
--- a/lisp/mastodon-toot.el
+++ b/lisp/mastodon-toot.el
@@ -146,8 +146,7 @@ Set `mastodon-toot--content-warning' to nil."
      "=================================================================\n"
      "Compose a new toot here. The following keybindings are available:"
      (mastodon-toot--format-kbinds kbinds)
-     "\n=================================================================\n"
-     "Toot Text:\n\n")))
+     "\n=================================================================\n\n")))
 
 (defun mastodon-toot--display-docs ()
   "Display documentation about mastodon-toot mode."
@@ -171,8 +170,7 @@ If REPLY-TO-ID is provided, set the MASTODON-TOOT--REPLY-TO-ID var."
     (switch-to-buffer-other-window buffer)
     (mastodon-toot--display-docs)
     (mastodon-toot--setup-as-reply reply-to-user reply-to-id)
-    (mastodon-toot-mode t)
-    ))
+    (mastodon-toot-mode t)))
 
 (defvar mastodon-toot-mode-map
   (let ((map (make-sparse-keymap)))
--- a/lisp/mastodon-toot.el
+++ b/lisp/mastodon-toot.el
@@ -74,8 +74,7 @@ Set `mastodon-toot--content-warning' to nil."
   (let ((re "^|=+=|$"))
     (save-excursion
       (goto-char 0)
-      (re-search-forward re)
-      (re-search-forward re) ; end of the docs
+      (re-search-forward re nil nil 2)
       (buffer-substring (+ 2 (point)) (+ 1 (length (buffer-string)))))))
 
 (defun mastodon-toot--send ()
--- a/lisp/mastodon-toot.el
+++ b/lisp/mastodon-toot.el
@@ -170,10 +170,12 @@ If REPLY-TO-ID is provided, set the MASTODON-TOOT--REPLY-TO-ID var."
   "Create a new buffer to capture text for a new toot.
 If REPLY-TO-USER is provided, inject their handle into the message.
 If REPLY-TO-ID is provided, set the MASTODON-TOOT--REPLY-TO-ID var."
-  (let ((buffer (get-buffer-create "*new toot*")))
+  (let* ((buffer-exists (get-buffer "*new toot*"))
+	 (buffer (or buffer-exists (get-buffer-create "*new toot*"))))
     (switch-to-buffer-other-window buffer)
-    (mastodon-toot--display-docs)
-    (mastodon-toot--setup-as-reply reply-to-user reply-to-id)
+    (when (not buffer-exists)
+      (mastodon-toot--display-docs)
+      (mastodon-toot--setup-as-reply reply-to-user reply-to-id))
     (mastodon-toot-mode t)))
 
 (defvar mastodon-toot-mode-map
--- a/lisp/mastodon-toot.el
+++ b/lisp/mastodon-toot.el
@@ -76,7 +76,7 @@ Set `mastodon-toot--content-warning' to nil."
       (goto-char 0)
       (re-search-forward re)
       (re-search-forward re) ; end of the docs
-      (buffer-substring (+ 2 (point)) (length (buffer-string))))))
+      (buffer-substring (+ 2 (point)) (+ 1 (length (buffer-string)))))))
 
 (defun mastodon-toot--send ()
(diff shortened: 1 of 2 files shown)


clarkenciel:
i realized a screenshot would be nice:
selection_002

lisp/mastodon-toot.el
```diff
@@ -118,6 +123,57 @@ Set `mastodon-toot--content-warning' to nil."
...
+
+(defun mastodon-toot--make-mode-docs ()
+  "Create formatted documentation text for the mastodon-toot-mode."
+  (let ((kbinds (mastodon-toot--get-mode-kbinds)))
+    (concat
+     "=================================================================\n"
+     "Compose a new toot here. The following keybindings are available:"
+     (mastodon-toot--format-kbinds kbinds)
+     "\n=================================================================\n"
+     "Toot Text:\n\n")))
```
jdenen: I don't think `Toot Text:` is necessary. It's obvious that everything under the `===` is content.

lisp/mastodon-toot.el
```diff
@@ -118,6 +123,57 @@ Set `mastodon-toot--content-warning' to nil."
...
+  (string-join (cons "" (mapcar #'mastodon-toot--format-kbind kbinds))
+	       "\n"))
+
+(defun mastodon-toot--make-mode-docs ()
+  "Create formatted documentation text for the mastodon-toot-mode."
+  (let ((kbinds (mastodon-toot--get-mode-kbinds)))
+    (concat
+     "=================================================================\n"
+     "Compose a new toot here. The following keybindings are available:"
+     (mastodon-toot--format-kbinds kbinds)
```
jdenen: This is a clever way to do it, but we might prefer a hard coded list. I only say that because I want to show the entire keybinding (`C-c C-c` vs `C-c`) and I want a description of behavior rather than the function name.

`mastodon-toot--toggle-warning` => Add/remove content warning
`mastodon-toot--send` => Send toot
`mastodon-toot--cancel` => Cancel toot

I'm not in love with the hard coded solution. If we can do these things dynamically, go for it.

lisp/mastodon-toot.el
```diff
@@ -118,6 +123,57 @@ Set `mastodon-toot--content-warning' to nil."
...
+(defun mastodon-toot--compose-buffer (reply-to-user reply-to-id)
+  "Create a new buffer to capture text for a new toot.
+If REPLY-TO-USER is provided, inject their handle into the message.
+If REPLY-TO-ID is provided, set the MASTODON-TOOT--REPLY-TO-ID var."
+  (let ((buffer (get-buffer-create "*new toot*")))
+    (switch-to-buffer-other-window buffer)
+    (mastodon-toot--display-docs)
+    (mastodon-toot--setup-as-reply reply-to-user reply-to-id)
+    (mastodon-toot-mode t)
+    ))
```
jdenen: Could you move these dangling `))` up onto the line above? Thanks.

lisp/mastodon.el
```diff
@@ -82,13 +82,8 @@
...
   (interactive)
-  (require 'mastodon-toot nil t)
-  (progn
-    (switch-to-buffer-other-window (get-buffer-create "*new toot*"))
-    (when user
-      (insert (format "@%s " user))
-      (setq mastodon-toot--reply-to-id reply-to-id))
-    (mastodon-toot-mode t)))
+  (require 'mastodon-toot nil t)  
+  (mastodon-toot--compose-buffer user reply-to-id))
```
jdenen, line 86: This is a much better looking function now. 👍

jdenen (changes requested):

lisp/mastodon-toot.el
clarkenciel: I think there's a way to do the keybindings dynamically, but the documentation big would be tricky dynamically. would it be acceptable to have something like:

`C-c C-c Send`
`C-c C-k Cancel`
`C-c C-w Toggle Warning`

that would involve manipulating the function's symbol name? or do you think that would just make the module too opinionated down the road?

i'm fine hard-coding, but i would love to have it "magically" know down the line.

jdenen: I'm OK with those. We'll just be cognizant of the approach when we name future modifiers.

alexjgriffith:
In the future could we set the text to a custom?

lisp/mastodon-toot.el
clarkenciel: great! i'll make the changes then.

clarkenciel:
@alexjgriffith probably, but which text in particular?

jdenen:
@alexjgriffith My only concern with customizing it is how that may impact #15 if that solution is going to live in the documentation alongside this string.

But we can figure all that out later. I'd want to get an impl for #15 before we decide anything.

clarkenciel:
okay @jdenen here's what it looks like now:
selection_006

lisp/mastodon-toot.el
```diff
@@ -69,10 +69,19 @@ Set `mastodon-toot--content-warning' to nil."
   (interactive)
   (mastodon-toot--kill))
 
+(defun mastodon-toot--remove-docs ()
+  "Get the body of a toot from the current compose buffer."
+  (let ((re "^|=+=|$"))
```
jdenen, line 74: I haven't cherry-picked and tried this out yet, but I'm surprised you don't need to escape those pipes.

```elisp
"^\|=+=\|$"
```

lisp/mastodon-toot.el
```diff
@@ -69,10 +69,19 @@ Set `mastodon-toot--content-warning' to nil."
...
+    (save-excursion
+      (goto-char 0)
+      (re-search-forward re)
+      (re-search-forward re) ; end of the docs
```
jdenen: No need to call twice. The last argument on `re-search-forward` is count. 

So:

```elisp
(re-search-forward re nil nil 2)
```
is like calling it twice.

jdenen (changes requested):

lisp/mastodon-toot.el
clarkenciel: I'm trying to match the literal `|` character and in emacs regexp that's just an unescaped `|` afaict. the `\|` is the regexp "or" character afaict:

given: `|=+=|`
`|` -> 2 matches
`\|` -> 5 matches (but no highlights, undefined behvavior i guess?)
`=\|+` -> 3 matches

lisp/mastodon-toot.el
clarkenciel: oh duh thanks!

clarkenciel:
@alexjgriffith i wonder if, for your "header replacement" suggestion we could make a custom function that constructs the contents of the header? then if you have want something like "a different koan every time i toot" you could have it pretty easily? given the stuff i'm working on for the counter, i may pull something like this into its own PR though since a custom header would require rethinking how i clean the header and counter out of the toot itself.

lisp/mastodon-toot.el
jdenen: Huh, TIL.

alexjgriffith:
@clarkenciel that would be cool, but I agree it can be its own feature sometime in the future.

jdenen (approved):

PR #98, opened 2017-04-27
Add the time of posting to the byline.

hdurer:
I was missing this when looking at a timeline.

Ideally we'll improve on that to show a relative time ("2 mins ago", "2 days ago", ... but that will require a timer that periodically updates the representation.

### DIFF
--- a/lisp/mastodon-tl.el
+++ b/lisp/mastodon-tl.el
@@ -31,6 +31,7 @@
 (require 'mastodon-http)
 (require 'mastodon-toot)
 (require 'mastodon-media)
+(require 'time-date)
 
 (defgroup mastodon-tl nil
   "Timelines in Mastodon."
@@ -130,6 +131,7 @@ Return value from boosted content if available."
 (defun mastodon-tl--byline (toot)
   "Generate byline for TOOT."
   (let ((id (cdr (assoc 'id toot)))
+        (timestamp (mastodon-tl--field 'created_at toot))
         (faved (mastodon-tl--field 'favourited toot))
         (boosted (mastodon-tl--field 'reblogged toot)))
     (propertize
@@ -140,6 +142,8 @@ Return value from boosted content if available."
                (format "(%s) " (propertize "F" 'face 'success)))
              (mastodon-tl--byline-author toot)
              (mastodon-tl--byline-boosted toot)
+             " "
+             (format-time-string mastodon-toot-timestamp-format (date-to-time timestamp))
              (propertize "\n  ------------" 'face 'default))
      'favourited-p faved
      'boosted-p    boosted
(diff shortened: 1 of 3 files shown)


lisp/mastodon-tl.el
```diff
@@ -130,6 +131,7 @@ Return value from boosted content if available."
 (defun mastodon-tl--byline (toot)
   "Generate byline for TOOT."
   (let ((id (cdr (assoc 'id toot)))
+        (timestamp (cdr (assoc 'created_at toot)))
```
jdenen: When the web client displays the timestamp for a boosted toot, does it show the time of the boost or the time of initial creation?

If it's the time of the boost, this will work just fine. If it's the time of initial creation, you'll want to use `mastodon-tl--field` so it grabs the boosted field.

test/mastodon-tl-tests.el
```diff
@@ -16,3 +44,56 @@
...
+  (let* ((toot `((favourited . t) (reblogged . t) ,@mastodon-tl-test-base-toot))
+         (timestamp (cdr (assoc 'created_at toot))))
+    (with-mock
+      (mock (date-to-time timestamp) => '(22782 21551))
+      (mock (format-time-string mastodon-toot-timestamp-format '(22782 21551)) => "2999-99-99 00:11:22")
+
+      (should (string= (substring-no-properties (mastodon-tl--byline toot))
+                       "
+ | (B) (F) Account 42 (@acct42@example.space) posted: 2999-99-99 00:11:22
+  ------------")))))
```
jdenen: Unit tests. They're beautiful. 😂

lisp/mastodon-tl.el
```diff
@@ -140,6 +142,8 @@ Return value from boosted content if available."
                (format "(%s) " (propertize "F" 'face 'success)))
              (mastodon-tl--byline-author toot)
              (mastodon-tl--byline-boosted toot)
+             " posted: "
```
jdenen: I don't think `posted:` is necessary. Appending the timestamp is enough.

I think

```
| Orbfix (@orbfix@mstdn.io) 2017-04-27 15:59:04
 ------------
```

looks better (less cluttered) than

```
| Orbfix (@orbfix@mstdn.io) posted: 2017-04-27 15:59:04
 ------------
```

jdenen (changes requested):

hdurer:
PR updated (also rebased to latest `develop` state).

 - `posted: ` removed
 - now using the timestamp of the reblogged toot not of the reblogging toot (that's what the web interface seems to do).
 - unit tests added for the reblogging case

jdenen (approved):

jdenen:
Looks great. Thanks @hdurer!

PR #113, opened 2017-05-02
Re-do the "Media_Links::" processing.

hdurer:
Re-do it as a loop rather than raising an error when the search fails and trying to catch that wholesale.

### DIFF
--- a/lisp/mastodon-media.el
+++ b/lisp/mastodon-media.el
@@ -42,23 +42,24 @@
   "Takes a URL and return an image."
   (let ((buffer (url-retrieve-synchronously url)))
     (unwind-protect
-	 (let ((data (with-current-buffer buffer
-		       (goto-char (point-min))
-		       (search-forward "\n\n")
-		       (buffer-substring (point) (point-max)))))
-	   (insert "\n")
-	   (insert-image (create-image data nil t)))
+        (let ((data (with-current-buffer buffer
+                      (goto-char (point-min))
+                      (search-forward "\n\n")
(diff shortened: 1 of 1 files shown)


lisp/mastodon-media.el
```diff
@@ -42,23 +42,24 @@
...
 
-contains `Media_Links::'."
-  (when start (goto-char (point-min)))
-  (search-forward-regexp "Media_Link::" nil nil nil)
-  (let ((start (progn (move-beginning-of-line '()) (point)))
-	(end (progn (move-end-of-line '()) (point))))
-    (list start end)))
+Returns the cons of (`start' . `end') points of that line or nil no
+more media links were found."
+  (let ((found? (search-forward-regexp "Media_Link::" nil t)))
```
jdenen: Change `found?` to `foundp` to match convention.

https://github.com/bbatsov/emacs-lisp-style-guide#naming

jdenen (changes requested):

lisp/mastodon-media.el
jdenen: This is a small nit but I like to follow convention. Otherwise, the PR looks great! Thanks.

hdurer:
Renamed and rebased to latest `develop`.

(Although technically, the style guide only prescribed the name of predicate methods not of boolean values. ☺ )

jdenen:
That's fair. The question mark looked weird in elisp though. 😛

jdenen (approved):

Issue #102, opened 2017-04-29
Fix line wrap issues in timeline


alexjgriffith:
Its a small thing, but for readability it would be nice if the toot fold width was adjustable or all regions that are not separated by two new lines were placed on one line and with visual-line-mode being an option to make it more ascetic.

alexjgriffith:
Closed via #114

PR #114, opened 2017-05-03
Closing #102 Fixing width

alexjgriffith:
# Background
`shr-render-region` by default outputs proportional font and creates a new line whenever the text reaches the buffer width. By using `customize-set-value` and setting `shr-use-fonts` to `nil` we have it output the width of the current buffer using `'default` monospaced text. Alternatively we can provide the max width used by setting the customize able value `shr-width` to our preferred width.

# Implementation
In `mastodon-tl.el` before any call to `shr-render-region` I have set the `shr-use-fonts` to nil

As a stop gap measure you can set the `shr-use-fonts` to nil globaly using the `customize-set-variables` function

## Before
screenshot 2017-05-02 22 56 44

## After
screenshot 2017-05-02 22 54 35

### DIFF
--- a/lisp/mastodon-tl.el
+++ b/lisp/mastodon-tl.el
@@ -160,7 +160,9 @@ also render the html"
    (with-temp-buffer
      (insert (decode-coding-string string 'utf-8))
      (when render
-       (shr-render-region (point-min) (point-max)))
+       (progn
+         (customize-set-value 'shr-use-fonts nil)
+         (shr-render-region (point-min) (point-max))))
      (buffer-string))
    'face face))
 
@@ -195,6 +197,7 @@ also render the html"
   (let ((content (mastodon-tl--field 'content toot)))
     (propertize (with-temp-buffer
                   (insert (decode-coding-string content 'utf-8))
+                  (customize-set-value 'shr-use-fonts nil)
                   (shr-render-region (point-min) (point-max))
                   (buffer-string))
                 'face 'default)))
--- a/lisp/mastodon-tl.el
+++ b/lisp/mastodon-tl.el
@@ -193,11 +193,11 @@ also render the html"
 
 (defun mastodon-tl--content (toot)
   "Retrieve text content from TOOT."
-  (let ((content (mastodon-tl--field 'content toot)))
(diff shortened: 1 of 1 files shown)


lisp/mastodon-tl.el
```diff
@@ -160,7 +160,9 @@ also render the html"
    (with-temp-buffer
      (insert (decode-coding-string string 'utf-8))
      (when render
-       (shr-render-region (point-min) (point-max)))
+       (progn
+         (customize-set-value 'shr-use-fonts nil)
```
jdenen: Any reason to use `customize-set-value` over `let`?

```elisp
(when render
  (let ((shr-render-region nil))
    (shr-render-region (point-min) (point-max))))
```

They appear to behave the same when opening up a timeline.

alexjgriffith: Yep, `let` works fine. Making the switch.

lisp/mastodon-tl.el
```diff
@@ -197,8 +196,8 @@ also render the html"
   (let ((content (mastodon-tl--field 'content toot)))
     (propertize (with-temp-buffer
                   (insert (decode-coding-string content 'utf-8))
-                  (customize-set-value 'shr-use-fonts nil)
-                  (shr-render-region (point-min) (point-max))
+                  (let ((shr-use-fonts nil))
```
jdenen: No need to call `let` again since it happens earlier in the function.

```elisp
(let ((content (mastodon-tl--field 'content toot))
      (shr-use-fonts nil))
  ;; ...
```

jdenen (changes requested):

hdurer:
Have you considered factoring out the html decoding logic into a helper function?

I expect in the future we'll want to do even more tweaking (e.g. detecting #tags and redirecting them to an Emacs buffer rather than leave them as an url link).

alexjgriffith:
@hdurer, i agree that as the html parsing logic is increased spinning off a separate function makes a lot of sense. At the same time the way we render toots needs a bit of a refactor to make it easier to support different types of notifications / statuses. Perhaps we could setup another rail as #84 is being finalized?

jdenen (approved):

PR #118, opened 2017-05-05
added option to bound the search for the next media link

alexjgriffith:
For async image loading I would like to limit the range of images to update. 

This PR allows you to pass an optional variable `end` to `mastodon-media--select-next-media-line` which bounds the range of `search-forward-regexp`

### DIFF
--- a/lisp/mastodon-media.el
+++ b/lisp/mastodon-media.el
@@ -51,10 +51,12 @@
       (kill-buffer buffer))))
 
 (defun mastodon-media--select-next-media-line (&optional end)
-  "Find coordinates of a line that contains `Media_Links::'
+  "Find coordinates of a line that contains `Media_Links::'.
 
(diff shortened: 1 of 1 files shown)


lisp/mastodon-media.el
```diff
@@ -50,12 +50,12 @@
           (insert-image (create-image data nil t)))
       (kill-buffer buffer))))
 
-(defun mastodon-media--select-next-media-line ()
+(defun mastodon-media--select-next-media-line (&optional end)
   "Find coordinates of a line that contains `Media_Links::'
```
hdurer: Document the new arg in the docstring?

jdenen (changes requested):
Needs rebased, but otherwise looks ready to go.

alexjgriffith:
We have moved forward with an async image loading. This can be revisited at a later time.

PR #121, opened 2017-05-10
Async image loading and avatars

hdurer:
This adds avatars to users' names (can be switched off and defaults to off if Emacs is not compiled with imagemagic since we need that for scaling).

It also makes the image loading asynchronous so we don't block on all the fetches.

### DIFF
--- a/lisp/mastodon-media.el
+++ b/lisp/mastodon-media.el
@@ -43,25 +43,51 @@
   (image-type-available-p 'imagemagick)
   "A boolean value stating whether to show avatars in timelines.")
 
-(defun mastodon-media--image-from-url (url media-type)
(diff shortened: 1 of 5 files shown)


jdenen:
@alexjgriffith Is your Emacs initialized with imagemagick? Mine is not, so I'll need to re-install Emacs to validate. But if you have it, you can validate.

I *can* validate that everything is working as expected for users without imagemagick. Looks good there. 👍

alexjgriffith:
I can validate that the images load, get more and update toots work, as well as boosting and favouriting.
screenshot 2017-05-12 17 22 18

alexjgriffith (approved):
Looks good, there is full functionality when imagemagick is installed and does not try to show avatars when imagemagick is unavailable.

PR #122, opened 2017-05-11
Fix for #84

alexjgriffith:
This PR lets us use `mastodon-mode` commands to update timeline like buffers, such as `*mastodon-notifications*` and `*mastodon-profile-user*`, and give the current mastodon buffers more user friendly names in the future (`*mastodon-local*` vs `*mastodon-public?local=true*`.

# Current Version

We now create a mastodon buffer using `mastodon-tl--init`. We pass it  a `buffer-name` an `endpoint` and an `update-function.
```
 (mastodon-tl--init
   "home" "timelines/home" 'mastodon-tl--timeline))
```

`mastodon-tl--init` overwrites the global `mastodon-buffer-spec` with a local buffer variable. If the buffer is not set localy its value will be `nil`
``` elsip
(defvar mastodon-buffer-spec nil)
```

`mastodon-buffer-spec` is a local buffer plist containing the "parsed" `buffer-name`, `endpoint` and `update-function`. There are getters that can be used in an mastodon-buffer to access these elements. 

``` elisp
(make-local-variable 'mastodon-buffer-spec)
(setq mastodon-buffer-spec
         (buffer-name "home" endpoint "timelines/home"
                                update-function mastodon-tl--timeline)
```

# Previous version
For simplicity this method still relies on the buffer naming convention. This is a liability and probably should be changed in the future to our own buffer local variable.

The parsed buffer name is matched against a list of plists.  A buffer specific endpoint is then passed to `mastodon-tl--more-json` and a buffer specific update function is used in `mastodon-tl--more`, rather than calling `mastodon-tl--timelines` directly. This list of plists is currently contained within `mastodon-tl--get-endpoint` but could be moved to a global variable so each buffer type could append their own endpoints and update function.

## Example mastodon-buffer-endpoints
``` elisp
          '((buffer-name "home" endpoint "timelines/home"
                         update-function mastodon-tl--timeline)
            (buffer-name "local" endpoint "timelines/public?local=true"
                         update-function 'mastodon-tl--timeline)
            (buffer-name "federated" endpoint "timelines/public"
                         update-function 'mastodon-tl--timeline)
            (buffer-name "public?local=true"
                         endpoint "timelines/public?local=true"
                         update-function 'mastodon-tl--timeline)
            (buffer-name "public"
                         endpoint "timelines/public"
                         update-function 'mastodon-tl--timeline)
            (buffer-name "notifications"
                         endpoint "notifications"
                         update-function
                         mastodon-notifications--notifications))
```

side note, renaming the `*mastodon-public?local=true*` buffer to `*mastodon-local*` works like a charm!


lisp/mastodon-tl.el
```diff
@@ -212,22 +212,72 @@ also render the html"
...
+                 (unless (functionp
+                          (plist-get endpoint-plist 'update-function))
+                   (error
+                    (format "%s not a function"
+                            (plist-get endpoint-plist 'update-function))))
+                 (cond
+                  ((stringp (plist-get endpoint-plist 'endpoint))
+                   endpoint-plist)
+                  ((functionp (plist-get endpoint-plist 'endpoint))
+                   (plist-put end 'endpoint
```
hdurer: This still says `end` instead of `endpoint-plist`.

This strongly suggests that it is not covered by any of your tests.   ;-)

alexjgriffith: I.e. perhaps I should write tests :p

Edit: Fixed for now. Will add unit tests tomorrow

hdurer:
Looks fine although we should add proper tests for this functionality in the future.
I'd also prefer if we re-visited the idea of storing the relevant data in buffer-local vars rather than decrypt the buffer name and devine the details from the entrails we find there.

BTW: shouldn't this be based off `develop` rather than `master`? 

Update:  Hmmm.. I see that this supposedly merges into `develop` but when I check out your branch it seems to be missing all the new stuff from there.  Have you confirmed things still work if merging onto `develop` ?

alexjgriffith:
@hdurer just updated my develop branch and it is looking good. 

I agree that a local buffer variable would be the way to go, and right now it will be easy to swap out the buffer name for them. My first priority now is to write unit tests.

alexjgriffith:
Currently working on including the endpoint and update-function in a local buffer variable

lisp/mastodon-tl.el
```diff
@@ -216,22 +216,72 @@ also render the html"
...
+                  ((stringp (plist-get endpoint-plist 'endpoint))
+                   endpoint-plist)
+                  ((functionp (plist-get endpoint-plist 'endpoint))
+                   (plist-put endpoint-plist 'endpoint
+                              (funcall (plist-get endpoint-plist 'endpoint)
+                                       buffer-string)))
+                  (t (error
+                      (format "%s not a string or function"
+                              (plist-get endpoint-plist 'endpoint))))))))
+         (endpoint (remove-if 'null
```
hdurer: `remove-if` seems to be part of cl.el and the tests seem to fail because of that.

alexjgriffith: I will replace remove-if, but the builds only stopped passing once i added my own unit tests.

lisp/mastodon-tl.el
```diff
@@ -223,22 +221,50 @@ also render the html"
...
+				    (or buffer (current-buffer))))
+
+(defun mastodon-tl--get-endpoint (&optional buffer)
+  "Get the ENDPOINT stored in `mastodon-buffer-spec'"
+  (mastodon-tl--get-buffer-property 'endpoint (or buffer (current-buffer))))
+
+(defun mastodon-tl--buffer-name ()
+  "Get the BUFFER-NAME stored in `mastodon-buffer-spec'"
+  (mastodon-tl--get-buffer-property 'buffer-name
+				    (or buffer (current-buffer))))
```
hdurer: There is no variable `buffer` here.
Missing the `&optional buffer` above, I guess.

lisp/mastodon-tl.el
```diff
@@ -223,22 +221,50 @@ also render the html"
...
-(defun mastodon-tl--more-json (timeline id)
-  "Return JSON for TIMELINE before ID."
-  (let ((url (mastodon-http--api (concat "timelines/"
-                                         timeline
-                                         "?max_id="
-                                         (number-to-string id)))))
+(defun mastodon-tl--get-update-function (&optional buffer)
+  "Get the UPDATE-FUNCTION stored in `mastodon-buffer-spec'"
+  (mastodon-tl--get-buffer-property 'update-function
+				    (or buffer (current-buffer))))
```
hdurer: `mastodon-tl--get-buffer-property` already handles a null buffer so no need to to the `or` here as well.

lisp/mastodon-tl.el
```diff
@@ -281,39 +307,54 @@ Move forward (down) the timeline unless BACKWARD is non-nil."
...
         (let ((inhibit-read-only t))
           (goto-char (point-min))
-          (mastodon-tl--timeline json))))))
+          (funcall update-function json))))))
+
+(defun mastodon-tl--init(buffer-name endpoint update-function)
+  "Initialize BUFFER-NAME with timeline targeted by ENDPOINT.
 
-(defun mastodon-tl--get (timeline)
+UPDATE-FUNCTION is used to recive more toots."
```
hdurer: "receive"

lisp/mastodon-tl.el
```diff
@@ -281,39 +307,54 @@ Move forward (down) the timeline unless BACKWARD is non-nil."
...
+  (let ((buffer (mastodon-tl--get buffer-name endpoint update-function)))
+    (with-current-buffer buffer
+      (make-local-variable 'mastodon-buffer-spec)
+      (setq mastodon-buffer-spec
+            `(buffer-name ,buffer-name
+                          endpoint ,endpoint update-function
+                          ,update-function)))))
+
+(defun mastodon-tl--get (buffer-name endpoint update-function)
```
hdurer: That name really needs improving.

lisp/mastodon-tl.el
```diff
@@ -281,39 +307,54 @@ Move forward (down) the timeline unless BACKWARD is non-nil."
 (defun mastodon-tl--more ()
   "Append older toots to timeline."
   (interactive)
-  (let* ((point-before (point))
-         (tl (mastodon-tl--timeline-name))
-         (id (mastodon-tl--oldest-id))
-         (json (mastodon-tl--more-json tl id)))
-    (when json
-      (with-current-buffer (current-buffer)
+  (with-current-buffer (current-buffer)
```
hdurer: ?!?  
`(current-buffer)` *is* already the current buffer.

lisp/mastodon-tl.el
```diff
@@ -281,39 +307,54 @@ Move forward (down) the timeline unless BACKWARD is non-nil."
...
+           (update-function (mastodon-tl--get-update-function))
+           (id (mastodon-tl--oldest-id))
+           (json (mastodon-tl--more-json endpoint id)))
+      (when json
         (let ((inhibit-read-only t))
           (goto-char (point-max))
-          (mastodon-tl--timeline json)
+          (funcall update-function json)
           (goto-char point-before)
           (mastodon-tl--goto-next-toot))))))
```
hdurer: If the `update-function` won't provide us with more data, then this will be an infinite loop, right?  (`mastodon-tl--goto-next-toot` will notice that there is no more data and recurse.  Or will mastodon fail and not return any json in that case?)

Note: this is not a new issue but may become one more urgently as we use this function for other types of timelines that *can* easily run out of old data.

lisp/mastodon-tl.el
```diff
@@ -281,39 +307,54 @@ Move forward (down) the timeline unless BACKWARD is non-nil."
...
 
 (defun mastodon-tl--update ()
   "Update timeline with new toots."
   (interactive)
-  (let* ((tl (mastodon-tl--timeline-name))
-         (id (mastodon-tl--newest-id))
-         (json (mastodon-tl--updated-json tl id)))
-    (when json
-      (with-current-buffer (current-buffer)
+  (with-current-buffer (current-buffer)
```
hdurer: Again.

lisp/mastodon-tl.el
alexjgriffith: I think I am going to envelop `mastodon-tl--get` into `mastodon-tl--init` since it has no purpose on its own.

lisp/mastodon-tl.el
alexjgriffith: Good catch, I will remove the goto-next toot line at the end and leave the pointer in its previous location.

jdenen (commented):
When I try to open the home timeline, I get the following error:

```
Debugger entered--Lisp error: (void-function mastodon-media--get-media-link-rendering)
  (mastodon-media--get-media-link-rendering preview-url)
 ```

From the `mastodon-tl--media` function.

jdenen (approved):

PR #123, opened 2017-05-12
closed #79

### LINKED ISSUE #79: Image impacts `mastodon-tl--goto-next-toot`
If you favorite a post while on the by-line with an image, it will send the favorite properly to the server but place a (F) on the following toot. If you favorite the post above the image the (F) will appear beside the image rather than in front of the by-line.

This effect is still there even when only `Media_Link:: /files/small/missing.png` is present.

alexjgriffith:
#79 and the odd behavior with favouriting/boosting toots with images was caused by preventing `"\n\n\n | "` from occurring before the byline. Toots with media would insert a media line between the two newlines at the end of the toot content and the bar at the start of the byline.

Moving the two newlines below the media line solved this issue.


lisp/mastodon-tl.el
```diff
@@ -180,12 +180,17 @@ also render the html"
...
-     media-attachements "")))
+  (let* ((media-attachements (mastodon-tl--field 'media_attachments toot))
+         (media-string (mapconcat
+			(lambda (media-attachement)
+			  (let ((preview-url
+				 (cdr (assoc 'preview_url media-attachement))))
+			    (mastodon-media--get-media-link-rendering
+			     preview-url)))
+			media-attachements "")))
+    (if (not (equal media-string ""))
```
jdenen, line 194: I'd prefer `(unless)` to `(if (not))`.

jdenen (changes requested):
One nit, but otherwise this looks great.

lisp/mastodon-tl.el
alexjgriffith: Unless only accepts one branch and I want to return `""` on false rather than `nil`.

jdenen: Ah, I missed that. I'll approve and merge then.

jdenen (approved):

jdenen:
@alexjgriffith Looks like you need to rebase on develop before this can be merged.

PR #136, opened 2017-05-24
Use relative time description for posting timestamps and periodically update these to keep then up-to-date

hdurer:


lisp/mastodon-tl.el
```diff
@@ -48,11 +48,20 @@
 
 (defvar mastodon-tl--buffer-spec nil
   "A unique identifier and functions for each Mastodon buffer.")
+(make-variable-buffer-local 'mastodon-tl--buffer-spec)
```
alexjgriffith: Nice clean up

lisp/mastodon-tl.el
```diff
@@ -163,7 +237,11 @@ Return value from boosted content if available."
              (mastodon-tl--byline-author toot)
              (mastodon-tl--byline-boosted toot)
              " "
-             (format-time-string mastodon-toot-timestamp-format (date-to-time timestamp))
+             (let ((parsed-time (date-to-time timestamp)))
```
alexjgriffith: If we don't use `timestamp` outside of defining `parsed-time` could we combine this `let` with the first `let` in this function? i.e., replace 

```elisp
(timestamp (mastodon-tl--field 'created_at toot))
```

with

 ```elisp 
```

File: lisp/mastodon-tl.el [hunk 3]
```diff
@@ -356,14 +525,23 @@ UPDATE-FUNCTION is used to recieve more toots."
...
     (mastodon-mode)
     (with-current-buffer buffer
-      (make-local-variable 'mastodon-tl--buffer-spec)
       (setq mastodon-tl--buffer-spec
             `(buffer-name ,buffer-name
                           endpoint ,endpoint update-function
-                          ,update-function)))
+                          ,update-function)
+            mastodon-tl--timestamp-update-timer (run-at-time mastodon-tl--timestamp-next-update
+                                                             nil ;; don't repeat
```
alexjgriffith: Is their an easy way to make the auto update on/off configurable?

test/mastodon-tl-tests.el
```diff
@@ -1,4 +1,5 @@
 (require 'el-mock)
+(require 'cl-macs)
```
alexjgriffith, line 2: Dependent on `cl-lib`. I'm not sure what the policy is on including `cl-lib` in tests though, will have to be checked by @jdenen

lisp/mastodon-tl.el
```diff
@@ -146,6 +155,71 @@ Return value from boosted content if available."
   (or (cdr (assoc field (cdr (assoc 'reblog toot))))
       (cdr (assoc field toot))))
 
+(defun mastodon-tl--relative-time-details (timestamp &optional current-time)
+  "Returns cons of (descriptive string . next change) for the TIMESTAMP.
```
alexjgriffith, line 167: checkdoc suggests using the imperative return rather than returns. However, it looks like their are endemic issues with the docs in this file, so perhaps we can go through and fix all the doc issues in a separate PR.

lisp/mastodon-tl.el
```diff
@@ -146,6 +155,71 @@ Return value from boosted content if available."
...
+                (cons "52 weeks ago"
+                      (* 60 60 24 7 52))
+              (funcall regular-response seconds-difference (* 60 60 24 7) "week")))
+           ((< seconds-difference (* 1.5 60 60 24 365))
+            (cons "one year ago"
+                  (* 60 60 24 365 1.5))) ;; a year and a half
+           (t
+            (funcall regular-response seconds-difference (* 60 60 24 365.25) "year")))))
+    (cons (car relative-result)
+          (time-add timestamp (seconds-to-time (cdr relative-result))))))
```
alexjgriffith, line 219: These human readable time strings are the bomb!

alexjgriffith (commented):
The new time strings look great and work on both my mac and windows machine without an issue. Only one nitpick and a couple suggestions for customizations.

Could we make the update time (5min) a custom-var, or have an on/off custom for running updates in the background?

lisp/mastodon-tl.el
hdurer: Agreed.  Let's keep it separate.  It's a habbit I picked up at work.  3rd person is the correct Java style here.

test/mastodon-tl-tests.el
hdurer: I asked in the matrix channel... let's see what he thinks.

I personally don't think it's an issue.
As I understand it, use of the "undesirable" files is even considered palatable as long as they are only required for bytecompiling (i.e. macros are ok if they don't expand into something undesirable to be used at runtime).  Surely the same reasoning would apply for tests: we don't want to burden users to require these files for normal operation but for tests (or compiling) it's ok.

lisp/mastodon-tl.el
hdurer: Good point!  Done.

lisp/mastodon-tl.el
hdurer: I added a customizable var to switch it all off.

Note that the 5mins mentioned above is the upper limit for a refresh; we could up that now (I'm fairly confident my next-update calculation is working properly now)...

hdurer:
I made some of the suggested changes.

hdurer:
Anybody want to re-review this after the last changes?

lisp/mastodon-tl.el
```diff
@@ -163,7 +245,10 @@ Return value from boosted content if available."
              (mastodon-tl--byline-author toot)
              (mastodon-tl--byline-boosted toot)
              " "
-             (format-time-string mastodon-toot-timestamp-format (date-to-time timestamp))
+             (propertize
+              (format-time-string mastodon-toot-timestamp-format parsed-time)
+              'timestamp parsed-time
+              'display (mastodon-tl--relative-time-description parsed-time))
```
alexjgriffith: Could we use the non relative timestamp here when `mastodon-tl--enable-relative-timestamps` is nil?

``` emacs-lisp
    'display (if mastodon-tl--enable-relative-timestamps
                           (mastodon-tl--relative-time-description parsed-time)
                         parsed-time)
```

alexjgriffith:
Looks good, checked it out on my Mac and windows. My only suggestion is to use the absolute time rather than relative time when `mastodon-tl--enable-relative-timestamps` is nil.

alexjgriffith (changes requested):

alexjgriffith (approved):
Went ahead and made the one small change.

Works fine, relative time strings are amazing. Setting `mastodon-tl--enable-relative-timestamps` to nil now disables the periodic updates and reverts to the absolute strings. We can change this behaviour in the future if desired.

PR #143, opened 2017-11-02
Use auth-source-search for fetching and saving password

rrix:
This gives users the ability to save their password to either the gpg-encrypted ~/.authinfo.gpg or
~/.authinfo so that they don't have to provide username/password each time. It's a bit messy due to this function's caller needing a buffer; I'm open to rewriting this to have two functions, one that just builds the post params alist, which also provides a place for people to advise it away if they'd prefer to not use auth-source for some reason.

### DIFF
--- a/lisp/mastodon-auth.el
+++ b/lisp/mastodon-auth.el
@@ -30,6 +30,7 @@
 ;;; Code:
 
 (require 'plstore)
+(require 'auth-source)
 
 (autoload 'mastodon-client "mastodon-client")
 (autoload 'mastodon-http--post "mastodon-http")
@@ -45,15 +46,29 @@
 
 (defun mastodon-auth--generate-token ()
   "Make POST to generate auth token."
-  (mastodon-http--post
-   (concat mastodon-instance-url "/oauth/token")
-   `(("client_id" . ,(plist-get (mastodon-client) :client_id))
-     ("client_secret" . ,(plist-get (mastodon-client) :client_secret))
-     ("grant_type" . "password")
-     ("username" . ,(read-string "Email: "))
-     ("password" . ,(read-passwd "Password: "))
-     ("scope" . "read write follow"))
-   nil))
+  (let* ((auth-source-creation-prompts
+          '((user . "Enter email for %h: ")
+            (secret . "Password: ")))
+         (credentials-plist (nth 0 (auth-source-search
+                                    :create t
+                                    :host mastodon-instance-url
+                                    :port 443
+                                    :require '(:user :secret)))))
+    (prog1
+        (mastodon-http--post
+         (concat mastodon-instance-url "/oauth/token")
+         `(("client_id" . ,(plist-get (mastodon-client) :client_id))
+           ("client_secret" . ,(plist-get (mastodon-client) :client_secret))
+           ("grant_type" . "password")
+           ("username" . ,(plist-get credentials-plist :user))
+           ("password" . ,(let ((secret (plist-get credentials-plist :secret)))
+                            (if (functionp secret)
+                                (funcall secret)
+                              secret)))
+           ("scope" . "read write follow"))
+         nil)
+      (when (functionp (plist-get credentials-plist :save-function))
+        (funcall (plist-get credentials-plist :save-function))))))
 
 (defun mastodon-auth--get-token ()
   "Make auth token request and return JSON response."
--- a/test/mastodon-auth-tests.el
+++ b/test/mastodon-auth-tests.el
@@ -3,19 +3,22 @@
 (ert-deftest generate-token ()
   "Should make `mastdon-http--post' request to generate auth token."
   (with-mock
-    (let ((mastodon-instance-url "https://instance.url"))
-      (mock (mastodon-client) => '(:client_id "id" :client_secret "secret"))
-      (mock (read-string "Email: ") => "foo@bar.com")
-      (mock (read-passwd "Password: ") => "password")
-      (mock (mastodon-http--post "https://instance.url/oauth/token"
-                                 '(("client_id" . "id")
-                                   ("client_secret" . "secret")
-                                    ("grant_type" . "password")
-                                    ("username" . "foo@bar.com")
-                                    ("password" . "password")
-                                    ("scope" . "read write follow"))
-                                 nil))
-      (mastodon-auth--generate-token))))
+   (let ((mastodon-instance-url "https://instance.url"))
+     (mock (mastodon-client) => '(:client_id "id" :client_secret "secret"))
+     (mock (auth-source-search :create t
+                               :host "https://instance.url"
+                               :port 443
+                               :require '(:user :secret))
+           => '((:user "foo@bar.com" :secret (lambda () "password"))))
+     (mock (mastodon-http--post "https://instance.url/oauth/token"
+                                '(("client_id" . "id")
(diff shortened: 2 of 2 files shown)


rrix:
I don't think I triggered that CI failure?

lisp/mastodon-auth.el
```diff
@@ -45,15 +45,31 @@
...
-     ("username" . ,(read-string "Email: "))
-     ("password" . ,(read-passwd "Password: "))
-     ("scope" . "read write follow"))
-   nil))
+  (let* ((auth-source-creation-prompts
+          '((user . "Enter email for %h: ")
+            (secret . "Password: ")))
+         (auth (nth 0 (auth-source-search
+                       :create t
+                       :host "cybre.space"
```
hdurer: Erm.  Hard coding the host here won't work, right?  
Or rather: this is the hostname used for the lookup, so can we use `mastodon-instance-url` here or could we parse out the hostname from that?

That way, when you switch `mastodon-instance-url` you get a different set of auth details here.

lisp/mastodon-auth.el
```diff
@@ -45,15 +45,31 @@
...
+                       :port 443
+                       :require '(:user :secret)))))
+    (let ((buffer
```
hdurer: I personally think that 
```
(prog1
    (mastodon-http--post ...)
  (when (functionp (plist-get auth :save-function))
    (funcall (plist-get auth :save-function))))
```
would be more readable as it expresses your intention better

hdurer (commented):
Cool.  I didn't know about that function.

Have you run the tests locally to see if they have just rotted or if your change does in fact cause failures?
In a way, with all the mocking happening in the tests I am a bit surprised your change does not require tweaks to the at least one test.


Also note that this project has been quite inactive in the past half year.  Two of the main contributors haven't even been in the matrix channel for many weeks...

Github does offer me the option to approve this change, so we could try (once you made the changes and we got tests working) to do that.

lisp/mastodon-auth.el
rrix: oh wow that's embarassing, huh

lisp/mastodon-auth.el
rrix: TIL about `prog1` and i can go fix a bunch of my other elisp code. :)

rrix:
I'll figure out the tests issue and update this sometime early this week.

rrix:
I'm having similar trouble running the ert-runner on my local machine, it hangs after loading all the `el` files. I'll take some more time on this after work today if I can.

lisp/mastodon-auth.el
hdurer: ☺ Great.  There is also `prog2` (I have never had any use for that so far) and `progn` which I assume you know already.

lisp/mastodon-auth.el
```diff
@@ -50,7 +50,7 @@
             (secret . "Password: ")))
          (auth (nth 0 (auth-source-search
```
hdurer: Instead of `nth 0` we could use `car` I guess?

lisp/mastodon-auth.el
```diff
@@ -53,23 +53,21 @@
...
-              ("client_secret" . ,(plist-get (mastodon-client) :client_secret))
-              ("grant_type" . "password")
-              ("username" . ,(plist-get auth :user))
-              ("password" . ,(let ((secret (plist-get auth :secret)))
-                               (if (functionp secret)
-                                   (funcall secret)
-                                 secret)))
-              ("scope" . "read write follow"))
-            nil)))
+    (prog1
```
hdurer: Is there any chance that `auth-source-search` could return `nil` and therefore `auth` be nil as well?
If so, consider handling that more gracefully here since surely the post will fail with a `nil` username and password.

lisp/mastodon-auth.el
```diff
@@ -45,15 +45,31 @@
...
-     ("client_secret" . ,(plist-get (mastodon-client) :client_secret))
-     ("grant_type" . "password")
-     ("username" . ,(read-string "Email: "))
-     ("password" . ,(read-passwd "Password: "))
-     ("scope" . "read write follow"))
-   nil))
+  (let* ((auth-source-creation-prompts
+          '((user . "Enter email for %h: ")
+            (secret . "Password: ")))
+         (auth (nth 0 (auth-source-search
```
hdurer: Nit: `auth` feels like a poor name here.  "authentication-data"? "credentials-plist"?

hdurer (commented):
The test times out because it seems to be hanging somewhere (asking for user inputs?).  We need to mock the call to `auth-source-search`.
With this diff it works for me:

```
--- a/test/mastodon-auth-tests.el
+++ b/test/mastodon-auth-tests.el
@@ -5,8 +5,11 @@
   (with-mock
     (let ((mastodon-instance-url "https://instance.url"))
       (mock (mastodon-client) => '(:client_id "id" :client_secret "secret"))
-      (mock (read-string "Email: ") => "foo@bar.com")
-      (mock (read-passwd "Password: ") => "password")
+      (mock (auth-source-search :create t
+                                :host "https://instance.url"
+                                :port 443
+                                :require '(:user :secret))
+            => '((:user "foo@bar.com" :secret (lambda () "password"))))
       (mock (mastodon-http--post "https://instance.url/oauth/token"
                                  '(("client_id" . "id")
                                    ("client_secret" . "secret")
```

rrix:
>  We need to mock the call to auth-source-search.

can you tell I don't write ERT tests for my code? :| I'll pull in your patch, and update the variable name.

lisp/mastodon-auth.el
rrix: my understanding of reading auth-source.el is that it'll return a plist either from one of the backends, or prompt the user if `:create` is true; The only way to quit out would be keyboard-quit which would abort the entire login flow. This is at least the case for netrc backend, and I've never seen anyone customize auth-sources in the wild or provide a custom backend

hdurer:
Great.  This looks fine to me.

I personally would prefer if you would squash all these commits into one, now that it has been reviewed.  That makes for a tidier revision history.

Also, this should be against the devel branch not master: a) because that's where development happens and b) that's where I can do approvals and merges according to an old conversation with jdenen.

rrix:
I'm so bad at github, i have no idea why it decided to add those develop commits to my PR even after changing its base. should i close this and open a fresh one?

This is what my git tree looks like locally:

screenshot_2017-11-12_10-19-04

hdurer:
That is indeed interesting.  I am no expert with github either.

The actual change seems to be what we want but the commits tab show 3 commits from develop and a merge with master.  That doesn't sound right...
The screenshot looks sane to me, I don't see how that matches up with what github shows.

I'd say, fiddle a bit more but if it doesn't work we'll abandon this PR and you create a new one.

hdurer:
Also: feel free to hop onto the `#mastodon.el:matrix.org` channel on Matrix to discuss things more directly than comments on githup PRs.

rrix:
(github decided to do a merge from master to develop when i switched the target branch of this PR, what a "nice" feature)

alexjgriffith:
Works fine. I am a little concerned that by default it saves the users password in plaintext in their home folder. Is there a way to have it require encryption or warn the user when they use the file that it is not stored safely?

alexjgriffith:
Superseded by #181

Issue #150, opened 2018-02-24
wrong-type-argument numberp on loading more toots


alexjgriffith:
## Description
In both the master and develop branches there is an error when more toots are requested in a time-line. When the requested is made emacs raises an error stating that the id is a wrong type.

## Suggestion
Requires a fix in `mastodon-tl--more-json` and `mastodon-tl--updated-json`. Check if the id is a number, if so convert, if not leave as is.

## Debugger 
``` emacs-lisp
Debugger entered--Lisp error: (wrong-type-argument numberp "99578427739496396")
  number-to-string("99578427739496396")
  (concat endpoint (if (string-match-p "?" endpoint) "&" "?") "max_id=" (number-to-string id))
  (mastodon-http--api (concat endpoint (if (string-match-p "?" endpoint) "&" "?") "max_id=" (number-to-string id)))
  (let* ((url (mastodon-http--api (concat endpoint (if (string-match-p "?" endpoint) "&" "?") "max_id=" (number-to-string id))))) (mastodon-http--get-json url))
  mastodon-tl--more-json("timelines/home" "99578427739496396")
  (let* ((point-before (point)) (endpoint (mastodon-tl--get-endpoint)) (update-function (mastodon-tl--get-update-function)) (id (mastodon-tl--oldest-id)) (json (mastodon-tl--more-json endpoint id))) (if json (progn (let ((inhibit-read-only t)) (goto-char (point-max)) (funcall update-function json) (goto-char point-before)))))
  mastodon-tl--more()
  funcall(mastodon-tl--more)
  (if npos (if (not (get-text-property npos (quote toot-id))) (mastodon-tl--goto-toot-pos find-pos refresh npos) (goto-char npos)) (funcall refresh))
  (let* ((npos (funcall find-pos (or pos (point)) (quote toot-id) (current-buffer)))) (if npos (if (not (get-text-property npos (quote toot-id))) (mastodon-tl--goto-toot-pos find-pos refresh npos) (goto-char npos)) (funcall refresh)))
  mastodon-tl--goto-toot-pos(next-single-property-change mastodon-tl--more 4629)
  (if (not (get-text-property npos (quote toot-id))) (mastodon-tl--goto-toot-pos find-pos refresh npos) (goto-char npos))
  (if npos (if (not (get-text-property npos (quote toot-id))) (mastodon-tl--goto-toot-pos find-pos refresh npos) (goto-char npos)) (funcall refresh))
[...]

PR #153, opened 2018-02-27
explicit media toggle and small bug fixes

### LINKED ISSUE #152: toggle async media loading
## Issue
With async toot loading and http streams I am running into issues when media is loaded asynchronously.

## Suggestion
Currently `mastodon-tl--show-avatars-p` is automatically toggled based on whether or not the client has `imagemagick`.  For debugging purposes, add a similar toggle `mastodon-tl--display-media-p` that:
1. prevents media stubs from being put into toots
2. prevents the async loading of those images
3. prevents loading avatars

alexjgriffith:
## Summary
This PR closes #152  and extends the fix for #150 to all places where ids were being converted from numbers to strings. It also adds some small fixes to the behaviour of threads and fixes boosting and favouriting (unfavouriting and unboosting still don't work).

## Changes
1. We now kill the http get request buffer once JSON has been extracted. 
2. `mastodon-tl--as-string` was implemented and replaced any occurrence of `number-to-string` or `int-to-string`
3. Added variable `mastodon-tl--display-media-p`. By default it is `'t` but can be made a local buffer variable and set to `nil`. When `nil` rather than displaying the media it just provides a link `Media::<link>`
4. Fixed checking for faves and boosts, they should now render properly. The return from `json-read-from-string` for `nil` is `:json-false` which evaluates to `'t` in elisp.
5. Fixed the `format` string that gets printed when faving and boosting
6. Fixed `mastodon-tl--thread` updating and requesting and changed its behaviour such that it tries to open the original toot thread rather than the boosted thread. 
7. Added tests for both the new `mastodon-tl--as-string` function and the `mastodon-tl--toot-id` utility.

### DIFF
--- a/lisp/mastodon-http.el
+++ b/lisp/mastodon-http.el
@@ -107,6 +107,7 @@ Pass response buffer to CALLBACK function."
                   (decode-coding-string
                    (buffer-substring-no-properties (point) (point-max))
                    'utf-8)))
+             (kill-buffer)
              (json-read-from-string json-string)))))
     json-vector))
 
--- a/lisp/mastodon-tl.el
+++ b/lisp/mastodon-tl.el
@@ -62,6 +62,9 @@ keep the timestamps current as time progresses."
   (image-type-available-p 'imagemagick)
   "A boolean value stating whether to show avatars in timelines.")
 
+(defvar mastodon-tl--display-media-p t
+  "A boolean value stating whether to show media in timelines.")
+
 (defvar mastodon-tl--timestamp-next-update nil
   "The timestamp when the buffer should next be scanned to update the timestamps.")
 (make-variable-buffer-local 'mastodon-tl--timestamp-next-update)
@@ -138,7 +141,7 @@ Optionally start from POS."
          (name (cdr (assoc 'display_name account)))
          (avatar-url (cdr (assoc 'avatar account))))
     (concat
-     (when mastodon-tl--show-avatars-p
+     (when (and mastodon-tl--show-avatars-p mastodon-tl--display-media-p)
        (mastodon-media--get-avatar-rendering avatar-url))
      (propertize name 'face 'mastodon-display-name-face)
      (propertize (concat " (@"
@@ -232,8 +235,8 @@ TIME-STAMP is assumed to be in the past."
   "Generate byline for TOOT."
   (let ((id (cdr (assoc 'id toot)))
         (parsed-time (date-to-time (mastodon-tl--field 'created_at toot)))
-        (faved (mastodon-tl--field 'favourited toot))
-        (boosted (mastodon-tl--field 'reblogged toot)))
+        (faved (equal 't (mastodon-tl--field 'favourited toot)))
+        (boosted (equal 't (mastodon-tl--field 'reblogged toot))))
     (propertize
      (concat (propertize "\n | " 'face 'default)
              (when boosted
@@ -289,11 +292,14 @@ also render the html"
                         (lambda (media-attachement)
                           (let ((preview-url
                                  (cdr (assoc 'preview_url media-attachement))))
-                            (mastodon-media--get-media-link-rendering
-                             preview-url)))
+                            (if mastodon-tl--display-media-p
+                                (mastodon-media--get-media-link-rendering
+                                 preview-url)
+                              (concat "Media::" preview-url "\n"))))
                         media-attachements "")))
-    (if (not (equal media-string ""))
-        (concat "\n" media-string ) "")))
+    (if (not (and (not mastodon-tl--display-media-p)
+                  (equal media-string "")))
+        (concat "\n" media-string) "")))
 
 
 (defun mastodon-tl--content (toot)
@@ -324,7 +330,8 @@ also render the html"
   (goto-char (point-min))
   (while (search-forward "\n\n\n | " nil t)
     (replace-match "\n | "))
-  (mastodon-media--inline-images))
+  (when mastodon-tl--display-media-p
+    (mastodon-media--inline-images)))
 
 (defun mastodon-tl--get-update-function (&optional buffer)
   "Get the UPDATE-FUNCTION stored in `mastodon-tl--buffer-spec'"
@@ -354,9 +361,7 @@ also render the html"
                                        "&"
                                      "?")
                                    "max_id="
-                                   (if (numberp id )
-                                       (number-to-string id)
-                                     id)))))
+                                   (mastodon-tl--as-string id)))))
     (mastodon-http--get-json url)))
 
 ;; TODO
@@ -369,9 +374,7 @@ also render the html"
                                       "&"
                                     "?")
                                   "since_id="
-                                  (if (numberp id)
-                                      (number-to-string id)
-                                    id)))))
+                                  (mastodon-tl--as-string id)))))
     (mastodon-http--get-json url)))
 
 (defun mastodon-tl--property (prop &optional backward)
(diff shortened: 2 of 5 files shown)


alexjgriffith:
Merged commits and replaced `alist-get` with `(cdr (assoc))`.

lisp/mastodon-http.el
```diff
@@ -100,13 +100,14 @@ Pass response buffer to CALLBACK function."
 (defun mastodon-http--get-json (url)
   "Make GET request to URL. Return JSON response vector."
   (let ((json-vector
-         (with-current-buffer (mastodon-http--get url)
+         (with-current-bufer (mastodon-http--get url)
```
hdurer: Surely that's wrong.  Should stay `buffer`

lisp/mastodon-http.el
```diff
@@ -100,7 +100,7 @@ Pass response buffer to CALLBACK function."
 (defun mastodon-http--get-json (url)
   "Make GET request to URL. Return JSON response vector."
   (let ((json-vector
-         (with-current-bufer (mastodon-http--get url)
+         (with-current-buffer (mastodon-http--get url)
```
hdurer: Why not squash this commit with the previous one to keep the history sane and readable?

lisp/mastodon-tl.el
```diff
@@ -293,11 +293,14 @@ also render the html"
...
-                            (mastodon-media--get-media-link-rendering
-                             preview-url)))
-                        media-attachements "")))
-    (if (not (equal media-string ""))
-        (concat "\n" media-string ) "")))
+                            (if mastodon-tl--display-media-p
+                                (mastodon-media--get-media-link-rendering
+                                 preview-url)
+                              (concat "Media::" preview-url ))))
+                        media-attachements "\n")))
```
hdurer: Why do we re-add the ugly newlines between images?  That's only needed if `mastodon-tl--display-media-p` is nil, or not?

lisp/mastodon-tl.el
```diff
@@ -293,11 +293,14 @@ also render the html"
...
+    (if (not (equal media-string ""))
```
hdurer: use `string=` and only do this newline prepending if not displaying media.

lisp/mastodon-tl.el
```diff
@@ -317,8 +320,7 @@ also render the html"
     (mastodon-tl--spoiler toot)
     ;; remove two trailing newlines
     (substring (mastodon-tl--content toot) 0 -2)
-    (when mastodon-tl--display-media-p
-      (mastodon-tl--media toot))
+    (mastodon-tl--media toot)
```
hdurer: We have only just made this change and now it is being undone again.  Would it make sense to just squash this commit with the first one?

lisp/mastodon-tl.el
```diff
@@ -1,4 +1,4 @@
-;;; mastodon-tl.el --- HTTP request/response functions for mastodon.el  -*- lexical-binding: t -*-
```
hdurer: a) all the files seem to use 3 semicolons, why change it in only one?
b) this seems unrelated to this PR.

lisp/mastodon-tl.el
```diff
@@ -399,21 +399,35 @@ Move forward (down) the timeline unless BACKWARD is non-nil."
   (goto-char (point-max))
   (mastodon-tl--property 'toot-id t))
 
+(defun mastodon-tl--as-string(id)
+  (cond ((numberp id )
```
hdurer: Add doc string

lisp/mastodon-tl.el
```diff
@@ -399,21 +399,35 @@ Move forward (down) the timeline unless BACKWARD is non-nil."
   (goto-char (point-max))
   (mastodon-tl--property 'toot-id t))
 
+(defun mastodon-tl--as-string(id)
```
hdurer: `mastodon-tl--id-as-string`?

lisp/mastodon-tl.el
```diff
@@ -399,21 +399,35 @@ Move forward (down) the timeline unless BACKWARD is non-nil."
...
-  (let* ((id (number-to-string (mastodon-tl--property 'toot-id)))
+  (let* ((id (mastodon-tl--as-string (mastodon-tl--property 'toot-id)))
          (url (mastodon-http--api (format "statuses/%s/context" id)))
          (buffer (format "*mastodon-thread-%s*" id))
          (toot (mastodon-tl--property 'toot-json))
          (context (mastodon-http--get-json url)))
     (with-output-to-temp-buffer buffer
       (switch-to-buffer buffer)
+      (mastodon-mode)
+      (setq mastodon-tl--buffer-spec
```
hdurer, line 433: Nit: This change seems unrelated to what the commit message says...

lisp/mastodon-toot.el
```diff
@@ -99,7 +99,7 @@ Remove MARKER if RM is non-nil."
     (mastodon-toot--action action
                            (lambda ()
                              (mastodon-toot--action-success "F" remove)
-                             (message (format "%sd #%d" action id))))))
+                             (message (format "%s #%d" action id))))))
```
hdurer: Now that ids are strings, wouldn't the second formatting param also be `%s`?

lisp/mastodon-tl.el
```diff
@@ -408,10 +408,24 @@ Move forward (down) the timeline unless BACKWARD is non-nil."
...
+  "Find approproiate toot id for toot under point.
+
+If the toot has been boosted use the id found in the
+reblog portion of the toot. Otherwise, use the body of
+the toot. This is the same behaviour as the mastodon.social
+webapp"
+  (let* ((json (mastodon-tl--property 'toot-json))
+         (id (alist-get 'id json ))
+         (reblog (alist-get 'reblog json )))
+    (if reblog
```
hdurer: Should this be wrapped in the new "make it a string" helper function to ensure we return valid ids?

lisp/mastodon-tl.el
```diff
@@ -423,7 +437,7 @@ Move forward (down) the timeline unless BACKWARD is non-nil."
             `(buffer-name ,buffer
                           endpoint ,(format "statuses/%s/context" id)
                           update-function
-                          mastodon-tl--timeline))
+                          (lambda(toot) (message "END of thread."))))
```
hdurer: Nit: space after `lambda`

lisp/mastodon-tl.el
```diff
@@ -236,8 +235,8 @@ TIME-STAMP is assumed to be in the past."
   "Generate byline for TOOT."
   (let ((id (cdr (assoc 'id toot)))
         (parsed-time (date-to-time (mastodon-tl--field 'created_at toot)))
-        (faved (equal ":json-true" (mastodon-tl--field 'favourited toot)))
-        (boosted (equal ":json-true" (mastodon-tl--field 'reblogged toot))))
+        (faved (equal 't (mastodon-tl--field 'favourited toot)))
```
hdurer: Whoops!  That looks like either our tests are wrong or some test is missing?

hdurer (commented):
First pass through these changes.  I find it very hard to read and understand this PR.  I think it would be preferable if you squashed all this into 2 or 3 commits.

There is really no need to keep all the false starts, accidental changes and undoing of those accidental changes.

lisp/mastodon-tl.el
alexjgriffith: This could be used anywhere we wish to convert a number to a string. To generalize I will rename the variable `ID` to `NUMERIC`.

lisp/mastodon-tl.el
alexjgriffith: Updated the commit message to include a line addressing this move. I was having odd behaviour in the thread buffer if i defined mastodon-tl--buffer-spec` before initializing `mastodon-mode`.

alexjgriffith:
Squashed the commits, for real this time, and generalized the variable names in `mastodon-tl--as-string`. This function should really be put in a utility module at some point.

lisp/mastodon-tl.el
hdurer: Ah ok, I now get it (reading the commit message).

Resolved.

lisp/mastodon-tl.el
hdurer: I still think we should not add newlines between images when we do show images...

lisp/mastodon-tl.el
hdurer: ok.

hdurer (dismissed):
Seems fine.  I suggest you still squash these two into one before merging into develop.

I still think adding newlines between images is wrong but I guess that's a matter of taste?

alexjgriffith:
Thanks for the quick feedback! We certainly do not need newlines between images. Will make a fix there and squash the commits.

hdurer (approved):
LGTM

Issue #158, opened 2018-03-02
Should `switch-to-buffer` be `with-current-buffer`?


alexjgriffith:
## Description
I noticed when streaming with media the active buffer will switch back to the mastodon buffer every time an image is loaded. This makes it impossible to do other tasks while steaming.

## Proposed Solution
I suggest switching out `switch-to-buffer` in `mastodon-media--process-image-response` with `with-current-buffer`

## Edit
Once this edit is made we can begin integrating `mastodon-async.el` into the main repo

hdurer:
Fixed with PR #160

PR #159, opened 2018-03-02
Closes issue #158

alexjgriffith:
- Small logic fix in `mastodon-tl--media` to remove redundant newline
- Replaced `switch-to-buffer` with `with-current-buffer` in `mastodon-media--process-image-response`


alexjgriffith:
Squashed merge with develop

PR #160, opened 2018-03-02
Fixes buffer switching issue and closes issue #158

alexjgriffith:
-Small logic fix in mastodon-tl--media to remove redundant newline
-Replaced switch-to-buffer with with-current-buffer in mastodon-media--process-image-response

Squashed with merge to develop


hdurer (approved):
Great!  Thanks for finding and fixing.

PR #161, opened 2018-03-02
Stop url.el from spamming the user with "Connecting host: ..." messages in the minibuffer

hdurer:
These happen during async loads, i.e. when the user might be doing something else completely.
Especially with auto-update that are coming this is becoming annoying.


alexjgriffith (approved):
LGTM

Ran on vanilla emacs with async and there were no extra request messages.

PR #163, opened 2018-03-02
Retoot add accts closes #155

### LINKED ISSUE #155: Replying to toot on another server fails to add the domain to the user handle
When you reply to a toot that originated on another server it does not add the server address to the handle. You  end up with `@` + `username` rather than `@` +`acct`. `acct` includes the domains, eg `@user` vs `@user@domain` . This is also an issue when you want to include more people in a response.  

## EDIT
Each status has a `mentions` which could automatically be pre-populated in a response toot when responding. Note there are `mentions` in both the status and the boosted status. They should be the same, but for ease of use we can put both in a list and insert a unique set into the reply.

alexjgriffith:
## Description
When responding to toots the full acct for both local and federated accounts are now added to the new toot buffer.

## Changes
- Added a function in mastodon.el to return the current user acct
- Added mastodon-toot--process-local, which takes an acct and appends the current server if it is local returns an empty string if the acct matches the current user and does only adds a prefix @ if the acct is federated
- mastodon-toot--mentions will return a formatted string of mentions or an empty string
- adds tests for mastodon-toot--mentions

## Edit (more changes)
- adds a missing , in mastodon-http--post
- `mastodon-toot--reply` now passes `mastodon-toot` a toot-id rather than the whole json
- 'mastodon-toot--reply-to-id is now a local var in a new toot
- only removes one trailing char rather than two from content in toot

### DIFF
--- a/lisp/mastodon-auth.el
+++ b/lisp/mastodon-auth.el
@@ -43,6 +43,9 @@
 (defvar mastodon-auth--token-alist nil
   "Alist of User access tokens keyed by instance url.")
 
+(defvar mastodon-auth--acct-alist nil
+  "Alist of account accts (name@domain) keyed by instance url.")
+
 (defun mastodon-auth--generate-token ()
   "Make POST to generate auth token."
   (mastodon-http--post
@@ -78,6 +81,21 @@ Generate token and set if none known yet."
         (setq token (plist-get json :access_token))
         (push (cons mastodon-instance-url token) mastodon-auth--token-alist)))
     token))
+
+(defun mastodon-auth--get-account-name ()
+  "Request user credentials and return an account name."
+  (cdr (assoc
+        'acct
+        (mastodon-http--get-json
+         (mastodon-http--api
+          "accounts/verify_credentials")))))
+
+(defun mastodon-auth--user-acct ()
+  "Return a mastodon user acct name."
+  (or (cdr (assoc  mastodon-instance-url mastodon-auth--acct-alist))
+      (let ((acct (mastodon-auth--get-account-name)))
+        (push (cons mastodon-instance-url acct) mastodon-auth--acct-alist)
+        acct)))
 
 (provide 'mastodon-auth)
 ;;; mastodon-auth.el ends here
--- a/lisp/mastodon-tl.el
+++ b/lisp/mastodon-tl.el
@@ -135,6 +135,7 @@ This also skips tab items in invisible text, i.e. hidden spoiler text."
         (search-pos (point)))
     (while (and (setq next-range (mastodon-tl--find-next-or-previous-property-range
                                   'mastodon-tab-stop search-pos nil))
+
                 (get-text-property (car next-range) 'invisible)
                 (setq search-pos (1+ (cdr next-range))))
       ;; do nothing, all the action in in the while condition
@@ -397,7 +398,6 @@ TIME-STAMP is assumed to be in the past."
                            (list 'invisible
                                  (not (get-text-property (car spoiler-text-region)
                                                          'invisible)))))))
-
 (defun mastodon-tl--make-link (string link-type)
   "Return a propertized version of STRING that will act like link.
 
@@ -488,6 +488,7 @@ message is a link which unhides/hides the main body."
   (insert
    (concat
     ;; remove trailing whitespace
+
     (replace-regexp-in-string
      "[\t\n ]*\\'" ""
      (if (mastodon-tl--has-spoiler toot)
--- a/lisp/mastodon-toot.el
+++ b/lisp/mastodon-toot.el
@@ -48,7 +48,6 @@
       map)
   "Keymap for `mastodon-toot'.")
 
-
 (defun mastodon-toot--action-success (marker &optional rm)
   "Insert MARKER with 'success face in byline.
 
@@ -142,14 +141,40 @@ Set `mastodon-toot--content-warning' to nil."
       (mastodon-http--triage response
                              (lambda () (message "Toot toot!"))))))
 
+(defun mastodon-toot--process-local (acct)
+  "Adds domain to local ACCT and replaces the curent user name with \"\".
+
+Mastodon requires the full user@domain, even in the case of local accts. 
+eg. \"user\" -> \"user@local.social \" (when local.social is the domain of the 
+mastodon-instance-url).
+eg. \"yourusername\" -> \"\"
+eg. \"feduser@fed.social\" -> \"feduser@fed.social\" "
+  (cond ((string-match-p "@" acct) (concat "@" acct " ")) ; federated acct
+        ((string= (mastodon-auth--user-acct) acct) "") ; your acct
+        (t (concat "@" acct "@" ; local acct
+                   (cadr (split-string mastodon-instance-url "/" t)) " "))))
+
+(defun mastodon-toot--mentions (status)
+  "Extract mentions from STATUS and process them into a string."
+  (interactive)
+  (let ((mentions (cdr (assoc 'mentions status))))
+    (mapconcat (lambda(x) (mastodon-toot--process-local
+                           (cdr (assoc 'acct x))))
+               ;; reverse does not work on vectors in 24.5
+               (reverse (append mentions nil))
+               "")))
+
 (defun mastodon-toot--reply ()
   "Reply to toot at `point'."
   (interactive)
   (let* ((toot (mastodon-tl--property 'toot-json))
          (id (mastodon-tl--as-string (mastodon-tl--field 'id toot)))
          (account (mastodon-tl--field 'account toot))
-         (user (cdr (assoc 'username account))))
-    (mastodon-toot user id)))
+         (user (cdr (assoc 'acct account)))
+         (mentions (mastodon-toot--mentions toot)))
+    (mastodon-toot (when user (concat (mastodon-toot--process-local user)
+                                      mentions))
+                   id)))
 
 (defun mastodon-toot--toggle-warning ()
   "Toggle `mastodon-toot--content-warning'."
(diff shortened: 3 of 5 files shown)


lisp/mastodon-toot.el
```diff
@@ -142,14 +141,53 @@ Set `mastodon-toot--content-warning' to nil."
...
   (let* ((toot (mastodon-tl--property 'toot-json))
          (id (mastodon-tl--as-string (mastodon-tl--field 'id toot)))
          (account (mastodon-tl--field 'account toot))
-         (user (cdr (assoc 'username account))))
-    (mastodon-toot user id)))
+         (user (cdr (assoc 'acct account)))
+         (mentions (mastodon-toot--mentions toot)))
+    (mastodon-toot (when user (concat (mastodon-toot--process-local user)
+                                      mentions))
+                   toot)))
```
hdurer: That second arg is meant to be an id not the whole toot.

lisp/mastodon-auth.el
```diff
@@ -43,6 +43,9 @@
 (defvar mastodon-auth--token-alist nil
   "Alist of User access tokens keyed by instance url.")
 
+(defvar mastodon-auth--acct-alist nil
+  "Alist of acct names keyed by instance url.")
```
hdurer: Is it just the account names or the whole alist of account details?

lisp/mastodon-toot.el
```diff
@@ -142,14 +141,53 @@ Set `mastodon-toot--content-warning' to nil."
       (mastodon-http--triage response
                              (lambda () (message "Toot toot!"))))))
 
+(defun mastodon-toot--get-acct()
```
hdurer: Unused function.

lisp/mastodon-toot.el
```diff
@@ -142,14 +141,53 @@ Set `mastodon-toot--content-warning' to nil."
...
+               (acct (concat "@" username "@" domain)))
+          (kill-new acct)
+          (message (format "Added %s to the kill ring." acct))
+          t)
+      (progn (message (format "No properly formated acct at point:%s." (point)))
+             nil))))
+
+(defun mastodon-toot--process-local (acct)
+  "Adds domain to local ACCT and replaces the curent user name with \"\"."
+  (cond ((string-match-p "@" acct) (concat "@" acct " "))
```
hdurer: Can you add comments, which cases you are dealing with here?

Also: why are we adding a white space?

lisp/mastodon-toot.el
```diff
@@ -142,14 +141,53 @@ Set `mastodon-toot--content-warning' to nil."
...
+               (username (string-remove-prefix "@" (car (last split-address))))
+               (acct (concat "@" username "@" domain)))
+          (kill-new acct)
+          (message (format "Added %s to the kill ring." acct))
+          t)
+      (progn (message (format "No properly formated acct at point:%s." (point)))
+             nil))))
+
+(defun mastodon-toot--process-local (acct)
+  "Adds domain to local ACCT and replaces the curent user name with \"\"."
```
hdurer: I don't understand this comment.

Does `acct` have to be a local name?  What is the "current user name"?

lisp/mastodon-toot.el
```diff
@@ -142,14 +141,53 @@ Set `mastodon-toot--content-warning' to nil."
...
+  (cond ((string-match-p "@" acct) (concat "@" acct " "))
+        ((string-match-p (mastodon-auth--user-acct) acct) "")
```
hdurer: Ah, is this removing the current user?

That seems orthogonal to what this function is meant to do.

Remove this and explicitly remove the current user from the list you produce with this?

hdurer: Also: use `string=` rather than `string-match-p`.

lisp/mastodon-toot.el
```diff
@@ -142,14 +141,53 @@ Set `mastodon-toot--content-warning' to nil."
...
+        (t (concat "@" acct "@"
+                   (cadr(split-string mastodon-instance-url "/" t)) " "))))
```
hdurer: nit: space after `cadr`.

lisp/mastodon-toot.el
```diff
@@ -142,14 +141,53 @@ Set `mastodon-toot--content-warning' to nil."
...
+
+(defun mastodon-toot--mentions (status)
+  "Extract mentions from STATUS and process them into a string."
+  (interactive)
+  (let ((mentions (cdr(assoc 'mentions status))))
```
hdurer: nit: space after `cdr`

lisp/mastodon-toot.el
```diff
@@ -142,14 +141,53 @@ Set `mastodon-toot--content-warning' to nil."
...
+    (if (> (length mentions) 0)
+        (mapconcat (lambda(x) (mastodon-toot--process-local
+                               (cdr(assoc 'acct x))))
+                   ;; reverse does not work on vectors in 24.5
+                   (reverse (mapcar (lambda(x) x) mentions)) "") "")))
```
hdurer: - There is a function `identity` which could replace your lambda
- The [manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Vector-Functions.html) mentions `(append avector nil)` as a way to convert a vector to a list.

hdurer: Put that third arg to `mapconcat` on its own line to make it clearer.

Also: that should probably be a space `" "` so that names are properly separated and you don't have to add the space in `mastodon-toot--process-local`.

test/mastodon-toot-tests.el
```diff
@@ -1,6 +1,46 @@
...
+               ((id . "1")
+                (username . "local")
+                (url . "")
+                (acct . "local"))])))
+
+(defconst mastodon-toot-no-mention
+  '((mentions . [])))
+
+(ert-deftest toot-multi-mentions ()
+  (should (string= (let ((mastodon-auth--acct-alist '(("https://local.social". "null")))
```
hdurer: I think this is more readable if you pull the `let` out.
That's basically the setup of the test context and now part of the actual test.

lisp/mastodon-toot.el
alexjgriffith: I add the spaces in `mastodon-toot-process local` to save us having to remove repeated spaces introduced from the null case (where your username is in the reply list).

alexjgriffith: Perfect, was looking for a way to do the conversion cleanly, this seemed unnecessary.

lisp/mastodon-auth.el
alexjgriffith: Just the account names

lisp/mastodon-toot.el
alexjgriffith: This function is meant to be used interactively by the user. Is there a specific way to mark such a function?

lisp/mastodon-toot.el
alexjgriffith: I add the white space here to account for the nil case. In mapconcat an extra space will be added.

lisp/mastodon-auth.el
```diff
@@ -79,5 +82,20 @@ Generate token and set if none known yet."
         (push (cons mastodon-instance-url token) mastodon-auth--token-alist)))
     token))
 
+(defun mastodon-auth--get-account ()
+  "Request user credentials and return an acct."
```
hdurer: Nit "account name"  Maybe even make that clear in the function name?

lisp/mastodon-tl.el
```diff
@@ -318,7 +318,7 @@ also render the html"
    (concat
     (mastodon-tl--spoiler toot)
     ;; remove two trailing newlines
-    (substring (mastodon-tl--content toot) 0 -2)
+    (substring (mastodon-tl--content toot) 0 -1)
```
hdurer: That comment above talks about 2 spaces being removed...

lisp/mastodon-toot.el
```diff
@@ -142,14 +141,61 @@ Set `mastodon-toot--content-warning' to nil."
...
+(defun mastodon-toot--process-local (acct)
+  "Adds domain to local ACCT and replaces the curent user name with \"\".
+
+Mastodon requires the full user@domain, even in the case of local accts. 
+eg. \"user\" -> \"user@local.social \" (when local.social is the domain of the 
+mastodon-instance-url).
+eg. \"yourusername\" -> \"\"
+eg. \"feduser@fed.social\" -> \"feduser@fed.social\""
+  (cond ((string-match-p "@" acct) (concat "@" acct " ")) ; federated acct
+        ((string-match-p (mastodon-auth--user-acct) acct) "") ; your acct
```
hdurer: Again: why not `string=` ?  Is there extra whitespace somewhere to be ignored?

lisp/mastodon-toot.el
```diff
@@ -142,14 +141,61 @@ Set `mastodon-toot--content-warning' to nil."
...
+        (t (concat "@" acct "@" ; local acct
+                   (cadr (split-string mastodon-instance-url "/" t)) " "))))
+
+(defun mastodon-toot--mentions (status)
+  "Extract mentions from STATUS and process them into a string."
+  (interactive)
+  (let ((mentions (cdr (assoc 'mentions status))))
+    (if (> (length mentions) 0)
```
hdurer: Nit: This check isn't needed, right?  If mentions is the empty vector, the mapappend will still do the right thing.

test/mastodon-toot-tests.el
```diff
@@ -1,6 +1,47 @@
...
+  (let ((mastodon-auth--acct-alist
+         '(("https://local.social". "null")))
+        (mastodon-instance-url "https://local.social"))    
+    (should (string= (mastodon-toot--mentions mastodon-toot-no-mention) ""))))
+
 (ert-deftest cancel ()
   (with-mock
     (mock (kill-buffer-and-window))
     (should (eq nil (mastodon-toot--cancel)))))
+
```
hdurer: Nit: remove the empty lines at end?

hdurer (commented):
Great.  Only a few nitty comments.

hdurer:
Oh.  About the function to copy the account name:  why not add it to the keymap so users have easy access to it (and document it in the README file)?

jdenen (approved):

Issue #166, opened 2018-03-03
Tab to next important feature


alexjgriffith:
@hdurer suggested the implementation of tab navigation (tab nav), whereby each time <tab> is pressed you are advanced to a important feature. 

Some possible features of interest would be
1. [x] hastags
2. [x] user handles
3. [x] byline (maybe?)
4. [x] weblinks
5. [x] content warnings

alexjgriffith:
Closed by #188

PR #168, opened 2018-03-03
Allow user to navigate interesting things in a buffer via tabbing.

hdurer:
Use tab to go forward, M-tab and S-Tab to go back.

This has always been possible while on a hyperlink but now works everywhere.

Currently only hyperlinks are tab stops but in the future we will want to support other things and there are already TODO comments in the code to note where we may want to do this.

This is the start of work #166 (Tab to next important feature).

### DIFF
--- a/lisp/mastodon-tl.el
+++ b/lisp/mastodon-tl.el
@@ -40,6 +40,7 @@
 (autoload 'mastodon-media--inline-images "mastodon-media")
 (autoload 'mastodon-mode "mastodon")
 (defvar mastodon-toot-timestamp-format)
+(defvar shr-use-fonts)  ;; need to declare it since Emacs24 didn't have this
 
 (defgroup mastodon-tl nil
   "Timelines in Mastodon."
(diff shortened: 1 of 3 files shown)


lisp/mastodon-tl.el
```diff
@@ -304,23 +373,17 @@ also render the html"
...
   "Display TOOT content and byline."
   (insert
    (concat
     (mastodon-tl--spoiler toot)
-    ;; remove two trailing newlines
-    (substring (mastodon-tl--content toot) 0 -2)
+    ;; remove trailing whitespace
+    (replace-regexp-in-string "[\t\n ]*\\'" "" (mastodon-tl--content toot))
     (mastodon-tl--media toot)
-    "\n\n"
```
alexjgriffith, line 323: This cleans things up nicely, however it breaks navigation via `mastodon-tl--goto-next-toot`, which relies on replacing "\n\n\n | " to override the text-properties on the leading portion of the byline. This was always a bit of a hack and eventually needed to be replaced.

We could rework `mastodon-tl--byline` such that "\n | " is outside of the outermost `propertize`. This would essentially do the same thing as  `(while (search-forward "\n\n\n | " nil t) (replace-match "\n | "))` in `mastodon-tl--timeline`.

lisp/mastodon-tl.el
```diff
@@ -73,6 +74,47 @@ keep the timestamps current as time progresses."
...
+  (let ((next-range (mastodon-tl--find-next-or-previous-property-range 'mastodon-tab-stop (point) nil)))
+    (if (null next-range)
+        (message "Nothing else here.")
+      (goto-char (car next-range))
+      (message "%s" (get-text-property (point) 'help-echo)))))
+
+(defun mastodon-tl--previous-tab-item ()
+  "Move to the previous interesting item.
+
+This could be the next toot, link, or image; whichever comes first.
```
alexjgriffith: *nit "previous"?

alexjgriffith (commented):
The patch looks good, nav works fine. 

Only one bug. Removing those extra newlines after media is added in `mastodon-tl--toot` messed up navigation using `j` and `k`. 

I think removing them is the way to go and had to be done eventually. I added my suggestion to fix this issue in the review.

## EDIT
Suggested changes, this is a simple change. We could do something more sophisticated with text properties as well.

``` emacs-lisp
(defun mastodon-tl--byline (toot)
  "Generate byline for TOOT."
        (faved (equal 't (mastodon-tl--field 'favourited toot)))
        (boosted (equal 't (mastodon-tl--field 'reblogged toot))))
    (concat "\n | " (propertize ; move "\n | " outside of propertize
                     (concat 
                      (when boosted
                        (format "(%s) "
                                (propertize "B" 'face 'mastodon-boost-fave-face)))
                      (when faved
                        (format "(%s) "
                                (propertize "F" 'face 'mastodon-boost-fave-face)))
                      " "
                      ;; TODO: Once we have a view for toot (responses etc.) make
                      ;; this a tab stop and attach an action.
                      (propertize
                       (format-time-string mastodon-toot-timestamp-format parsed-time)
                       'timestamp parsed-time
                       'display (if mastodon-tl--enable-relative-timestamps
                                    (mastodon-tl--relative-time-description parsed-time)
                                  parsed-time))
                      (propertize "\n  ------------" 'face 'default))
                     'favourited-p faved
                     'boosted-p    boosted
                     'toot-id      id
                     'toot-json    toot))))
=
```

lisp/mastodon-tl.el
```diff
@@ -260,26 +306,49 @@ TIME-STAMP is assumed to be in the past."
...
-     (when render
-       (let ((shr-use-fonts nil))
-         (shr-render-region (point-min) (point-max))))
-     (buffer-string))
-   'face face))
+(defun mastodon-tl--render-text (string)
+  "Returns a propertized text giving the rendering of the given HTML string."
+  (with-temp-buffer
+    (insert string)
+    (let ((shr-use-fonts t)
```
alexjgriffith: Can we make `shr-use-fonts` a flag? I'd much rather not use the shr fonts.

hdurer: Ah yes.  That shouldn't be here - has nothing to do with tabbed browsing.  I'll revert and re-add it gated by a var in a later change.

lisp/mastodon-tl.el
hdurer: Indeed.

lisp/mastodon-tl.el
hdurer: Agreed.  Let's do this better.  PTAL.

alexjgriffith (approved):
LGTM

PR #169, opened 2018-03-05
Add new customization flag to switch between fixed width / proportional fonts.

hdurer:
The default stays default-width as is now.


alexjgriffith (approved):
LGTM

PR #174, opened 2018-03-06
Generalizations for notifications

alexjgriffith:
Generalizations of `mastodon-tl--toot` and `mastodon-tl--byline` that allow them to be used to render notifications.

Three new arguments are added to `mastodon-tl--toot`
1. body: appears above the byline
2. author-byline, a functions that defines how the author should be rendered and that defaults to `mastodon-tl--byline-author`
2. action-byline, a functions that defines how the action should be rendered. Actions include boosting, following, mentioned etc. It defaults to `mastodon-tl--byline-boosted`

I also removed the redundant removal of newlines in `mastodon-tl--timeline`

### DIFF
--- a/lisp/mastodon-tl.el
+++ b/lisp/mastodon-tl.el
@@ -332,7 +332,7 @@ TIME-STAMP is assumed to be in the past."
   "Generate byline for TOOT.
 
 AUTHOR-BYLINE is function for adding the author portion of 
-the byline that takes one variable. By default it is `mastodon-tl--byline-author'
+the byline that takes one variable.
 ACTION-BYLINE is a function for adding an action, such as boosting
 favouriting and following to the byline. It also takes a single function. By default
 it is `mastodon-tl--byline-boosted'"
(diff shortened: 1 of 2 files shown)


lisp/mastodon-tl.el
```diff
@@ -324,8 +324,14 @@ E.g. this could return something like \"1 min ago\", \"yesterday\", etc.
 TIME-STAMP is assumed to be in the past."
   (car (mastodon-tl--relative-time-details timestamp current-time)))
 
-(defun mastodon-tl--byline (toot)
-  "Generate byline for TOOT."
+(defun mastodon-tl--byline (toot author-byline action-byline)
+  "Generate byline for TOOT.
+
+AUTHOR-BYLINE is function for adding the author portion of 
+the byline that takes one variable. By default it is `mastodon-tl--byline-author'
```
hdurer: `By default`is now meaningless.

lisp/mastodon-tl.el
```diff
@@ -483,26 +489,39 @@ message is a link which unhides/hides the main body."
      (mastodon-tl--render-text content)
      (mastodon-tl--media toot))))
 
-(defun mastodon-tl--toot (toot)
-  "Display TOOT content and byline."
+(defun mastodon-tl--insert (toot body author-byline action-byline)
```
hdurer: `mastodon-tl--insert` is too generic  name.   
If this going to be more generic than just inserting toots, we should come up with a good name for this generic "something" and then also rename the `toot` argument and fix the doc comment to not say `toot`.

lisp/mastodon-tl.el
```diff
@@ -483,26 +489,39 @@ message is a link which unhides/hides the main body."
...
+  "Display toot content and byline.
+
+BODY will form the section of the toot above the byline.
+AUTHOR-BYLINE and ACTION-BYLINE are optional functions that are passed to
```
hdurer: Remove the `optional` bits.

hdurer (approved):

PR #175, opened 2018-03-06
Fix missing thread notifications

alexjgriffith:
`mastodon-tl--thread` throws errors when used with notifications. 

This PR adds some defense arround `mastodon-tl--timeline` as called from within `mastodon-tl--thread`


alexjgriffith:
This has to be rebased on the most recent version of develop, will fix tomorrow.

PR #179, opened 2018-03-08
Clean up code to remove compilation warnings.

hdurer:
We used to be clean but have slipped recently.
Let's clean up the code so it's easier to see real code smell warnings in the compilation.


alexjgriffith (approved):
LGTM

PR #181, opened 2018-03-09
Optionally use auth-source-search for fetching and saving password.

hdurer:
This gives users the ability to save their password to either the gpg-encrypted ~/.authinfo.gpg or
~/.authinfo so that they don't have to provide username/password each time

By default this feature is off since without any changes, things would default to saving credentials to unencrypted storage which is not a good user default.

### DIFF
--- a/lisp/mastodon-auth.el
+++ b/lisp/mastodon-auth.el
@@ -43,6 +43,15 @@
   :prefix "mastodon-auth-"
   :group 'mastodon)
 
+(defcustom mastodon-auth-source-file ""
+  "Filename to use to store user names and passwords.
+
+Leave empty to not permanently store any secrets.
+Otherwise set to e.g. \"~/.authinfo.gpg\" to have encrypted storage, or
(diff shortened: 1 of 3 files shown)


lisp/mastodon-auth.el
```diff
@@ -61,6 +78,36 @@
    nil
    :unauthenticated))
 
+(defun mastodon-auth--generate-token-and-store ()
+  "Make POST to generate auth token.
+
+Reads and/or stores secres in `MASTODON-AUTH-SOURCE-FILE'."
+  (let* ((auth-sources (list mastodon-auth-source-file))
+	 (auth-source-creation-prompts
```
alexjgriffith, line 86: nit indentation? may be a tabs issue again. Will capture this when we do a project wide cleanup

alexjgriffith (approved):
LGTM. some indentation issues probably caused by tabs. These will be cleaned up in a project wide cleanup.

PR #182, opened 2018-03-10
Fixed jk nav issue and replaced tabs with spaces in mastodon-tl.el

alexjgriffith:
1. Added `byline` text property to toot byline
2. `mastodon-tl--goto-toot-pos` now searches for next `byline` rather than `toot-id`
3. Replaced tabs with spaces in mastodon-tl.el. Tabs look fine in emacs but don't seam to render properly in github. This should be done for all files at some point.


hdurer (approved):
Thanks

PR #183, opened 2018-03-10
Some small tweaks to our keymap handling.

hdurer:
- Use special mode which sets up a sensible base keymap.
- Define the keymap just once so that users can tweak it as they wish
  without us messing with that on each invocation of mastodon-mode.


alexjgriffith (approved):
LGTM. Note that `simple-mode` breaks threads. For some reason threads where never `read-only`. Will look into putting up a small patch for this.

PR #186, opened 2018-03-10
Add user profiles closes #165

### LINKED ISSUE #165: User Profiles
Add the ability to browse user profiles. Checkout  [mastodon-future.el/mastodon-profile](https://github.com/alexjgriffith/mastodon-future.el/blob/master/lisp/mastodon-profile.el)

- [x] load their toots, like a timeline
- [x] view their followers
- [x] view those who follow them

alexjgriffith:
## Changes
- Adds user profiles 
- also provides functions `mastodon-profile--get-following` and `mastodon-profile--get-followers`
- added keybinding U for opening user profiles
- autoloads `mastodon-profile--get-next-author`

## Notes
Should add tests for a few functions. But most of this builds directly on `mastodon-tl` functionality.

### DIFF
--- a/lisp/mastodon-profile.el
+++ b/lisp/mastodon-profile.el
@@ -1,0 +1,154 @@
+;;; mastodon-profile.el --- Functions for inspecting Mastodon profiles -*- lexical-binding: t -*-
+
+;; Copyright (C) 2017 Johnson Denen
+;; Author: Johnson Denen <johnson.denen@gmail.com>
+;; Version: 0.7.2
+;; Package-Requires: ((emacs "24.4"))
+;; Homepage: https://github.com/jdenen/mastodon.el
+
+;; This file is not part of GNU Emacs.
+
+;; This file is part of mastodon.el.
+
+;; mastodon.el is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; mastodon.el is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with mastodon.el.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; mastodon-profile.el generates stream of users toots.
+;; To fix
+;; 1. Render Image at top of frame [x]
+;; 2. Get toot author [x]
+;; 3. Load more toots [x]
+;; Later
+;; 1. List followers [x]
+;; 2. List people they follow [x]
+;; 3. Option to follow
+;; 4. wheather they follow you or not
+;; 5. Show only Media
+
+;;; Code:
+
+(defun mastodon-profile--toot-json ()
+  "Get the next toot-json."
+  (interactive)
+  (mastodon-tl--property 'toot-json))
+
+(defun mastodon-profile--make-author-buffer (status)
+  "Take a STATUS and inserts a user account into a new buffer."
+  (let* ((id (mastodon-profile--field status 'id))
+         (acct (mastodon-profile--field status 'acct))
+         accounts/id/statuses
+         (url (mastodon-http--api
+               (concat "accounts/"
+                       (format "%s" id)
+                       "/statuses" )))
+         (buffer (concat "*mastodon-" acct  "*"))
+         (account (cdr(assoc 'account status)))
+         (note (mastodon-profile--field status 'note))
+         (json (mastodon-http--get-json url)))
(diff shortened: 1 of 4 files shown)


lisp/mastodon-profile.el
```diff
@@ -0,0 +1,154 @@
...
+               (concat "accounts/"
+                       (format "%s" id)
+                       "/statuses" )))
+         (buffer (concat "*mastodon-" acct  "*"))
+         (account (cdr(assoc 'account status)))
+         (note (mastodon-profile--field status 'note))
+         (json (mastodon-http--get-json url)))
+    (with-output-to-temp-buffer buffer
+      (switch-to-buffer buffer)
+      (mastodon-mode)
```
hdurer: If you move this to the very end, once you have created all the contents then you can get rid of the `inhibit-read-only` setting below.

lisp/mastodon-profile.el
```diff
@@ -0,0 +1,154 @@
...
+      (setq mastodon-tl--buffer-spec
+            `(buffer-name ,buffer
+                          endpoint ,(format "accounts/%s/statuses" id)
+                          update-function
+                          ,'mastodon-tl--timeline json))
+      (let ((inhibit-read-only t))
+        "\n"
```
hdurer: You need to wrap all this until the `(mastodon-tl--timeline json)` below in an `insert` call.
On my machine this doesn't make it into the buffer otherwise.

lisp/mastodon-profile.el
```diff
@@ -0,0 +1,154 @@
...
+        (mastodon-profile--image-from-status account)
+        "\n"
+        (propertize (mastodon-profile--field status 'display_name
```
hdurer: This is bracketed incorrectly.  `(mastodon-profile--field status 'display_name)` is the inner expression otherwise I get an error and there are no properties to set.

lisp/mastodon-profile.el
```diff
@@ -0,0 +1,154 @@
...
+            (insert (propertize (mastodon-tl--byline-author
+                                 (list (append (list 'account) toot)))
+                                'byline  't
+                                'toot-id (cdr(assoc 'id toot)) 'toot-json toot)
+                    "\n"))
+          tootv))
+  (mastodon-media--inline-images))
+
+
+(defun mastodon-profile--get-following ()
```
hdurer, line 124: This and `mastodon-profile--followers` aren't hooked up yet?

lisp/mastodon-profile.el
```diff
@@ -0,0 +1,154 @@
...
+  (interactive)
+  (mastodon-profile--make-author-buffer (mastodon-profile--toot-json)))
+
+(defun mastodon-profile--image-from-status (status)
+  "Generate an image from a STATUS."
+  (let ((url (cdr(assoc 'avatar_static status))))
+    (unless (equal url "/avatars/original/missing.png")
+      (mastodon-media--get-media-link-rendering url))))
+
+(defun mastodon-profile--field (status field)
```
hdurer: Neither the function name nor the docstring gives me an idea of what this does.

I think you should work with the account json object and not the toot.  Then you can lose the inner assoc and this becomes a simple concatenation of `cdr` and `assoc` which doesn't really add any abstraction so we could get rid of it altogether.

lisp/mastodon-profile.el
```diff
@@ -0,0 +1,154 @@
...
+  (get-authour-id (toot-proporties)))
+
+(defun mastodon-profile--add-author-bylines (tootv)
+  "Convert TOOTV into a author-bylines and insert."
+  (let ((inhibit-read-only t))
+    (mapc (lambda(toot)
+            (insert (propertize (mastodon-tl--byline-author
+                                 (list (append (list 'account) toot)))
+                                'byline  't
+                                'toot-id (cdr(assoc 'id toot)) 'toot-json toot)
```
hdurer: Nit: Missing space after `cdr` here and twice more below.

hdurer (commented):
I really like this!

A few changes had to be made to make this work for me.

Thinking ahead how this can work given a generic user handle (when we process the "@userid" links in the body) I suspect we should work off account objects and not full toots?

Nothing in `mastodon-profile--make-author-buffer` really needs the full status but only wants data from the account sub-object.  (for that to be true, we need to make the second arg to `mastodon-tl--render-text` optional - it won't be needed when rendering a user's self description I suspect).

lisp/mastodon-profile.el
alexjgriffith: They work, but i could not think of an intuitive key binding.

We can brainstorm and give them a keybinding before the 0.8.0 release.

lisp/mastodon-profile.el
alexjgriffith: If I don't put `mastodon-mode` up top I get a weird error when requesting more toots stating that `mastodon-tl--buffer-spec` is not defined.

lisp/mastodon-profile.el
hdurer: Sounds good.

hdurer (approved):
Cool.  This is quite exciting!

PR #188, opened 2018-03-15
Make users handles tabstops

hdurer:
These currently are still only links to the web page with the profile.
Soon we will have native profile browsing and we can replace these links with calls to that functionality.


alexjgriffith (approved):
LGTM, works like a charm

PR #189, opened 2018-03-16
Adjusted mastodon-tl.el to open user profiles in emacs.

alexjgriffith:
This PR preps for opening user profiles within emacs.

## DESCRIPTION
Rather than opening a users profile in a browser, mastodon.el now searches for the user handle and returns an account or a nil. The account can be fed directly into `mastodon-profile--make-author-buffer` once PR #186 is merged.

### DIFF
--- a/lisp/mastodon-tl.el
+++ b/lisp/mastodon-tl.el
@@ -72,6 +72,12 @@ width fonts when rendering HTML text"))
   (image-type-available-p 'imagemagick)
   "A boolean value stating whether to show avatars in timelines.")
 
+(defvar mastodon-tl-update-point nil
+  "When updating a mastodon buffer this is where new toots will be inserted.
+
+If nil `(point-min)' is used instead.")
+(make-variable-buffer-local 'mastodon-tl-update-point)
+
 (defvar mastodon-tl--display-media-p t
   "A boolean value stating whether to show media in timelines.")
 
@@ -244,7 +250,8 @@ Optionally start from POS."
                  'face 'mastodon-handle-face
                  'mouse-face 'highlight
 		 ;; TODO: Replace url browsing with native profile viewing
-		 'mastodon-tab-stop 'handle
+		 'mastodon-tab-stop 'user-handle
+                 'account (cdr (assoc 'account toot))
 		 'shr-url profile-url
 		 'keymap mastodon-tl--link-keymap
                  'mastodon-handle (concat "@" handle)
@@ -415,12 +422,15 @@ links in the text. If TOOT is nil no parsing occurs."
 
           (;; User handles:
            maybe-userhandle
-           ;; Until we have native profile page viewing we leave these as HTML
-           ;; links.
-           (setq mastodon-tab-stop-type 'handle
-                 keymap mastodon-tl--link-keymap
-                 help-echo (concat "Browse user profile of " maybe-userhandle)
-                 extra-properties (list 'mastodon-handle maybe-userhandle)))
+           (let ((maybe-userid (mastodon-tl--extract-userid-toot
+                                toot maybe-userhandle)))
+             (setq mastodon-tab-stop-type 'user-handle
+                   keymap mastodon-tl--link-keymap
+                   help-echo (concat "Browse user profile of " maybe-userhandle)
+                   extra-properties (append
+                                     (list 'mastodon-handle maybe-userhandle)
+                                     (when maybe-userid
+                                       (list 'acccount-id maybe-userid))))))
           ;; Anything else:
           (t
            ;; Leave it as a url handled by shr.el.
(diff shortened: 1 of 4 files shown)


alexjgriffith:
A couple of errors it seems that there is an issue with the byline rendering. There are an extra two spaces in front of the user's display name. 

``` emacs-lisp
(ert-test-failed ((should (string= (substring-no-properties byline) "\n | Account 42 (@acct42@example.space) 2999-99-99 00:11:22\n  ------------")) :form (string= "\n |   Account 42 (@acct42@example.space) 2999-99-99 00:11:22\n  ------------" "\n | Account 42 (@acct42@example.space) 2999-99-99 00:11:22\n  ------------") :value nil))
```

lisp/mastodon-tl.el
```diff
@@ -244,9 +244,10 @@ Optionally start from POS."
                  'face 'mastodon-handle-face
                  'mouse-face 'highlight
 		 ;; TODO: Replace url browsing with native profile viewing
-		 'mastodon-tab-stop 'shr-url
+		 'mastodon-tab-stop 'handle
```
hdurer: `user-handle` or some thing like that?  "handle" is awfully generic.

lisp/mastodon-tl.el
```diff
@@ -244,9 +244,10 @@ Optionally start from POS."
...
 		 'shr-url profile-url
-		 'keymap mastodon-tl--shr-map-replacement
+		 'keymap mastodon-tl--link-keymap
+                 'mastodon-handle (concat "@" handle)
```
hdurer: Here we actually have the account info, why not store that in the property (or some other property)?
Seems wasteful (resource wasting for the server and time wasting for the user) to go back to the server to look up information we already have.

lisp/mastodon-tl.el
```diff
@@ -415,10 +416,10 @@ links in the text."
...
            ;; Until we have native profile page viewing we leave these as HTML
            ;; links.
-           (setq mastodon-tab-stop-type 'shr-url
-                 keymap mastodon-tl--shr-map-replacement
-                 help-echo (concat "Browse user profile of " maybe-userhandle)))
-
+           (setq mastodon-tab-stop-type 'handle
+                 keymap mastodon-tl--link-keymap
+                 help-echo (concat "Browse user profile of " maybe-userhandle)
+                 extra-properties (list 'mastodon-handle maybe-userhandle)))
```
hdurer: We can also look up the id in the toot's `mention` field.
With that we still have to go to the server to look up the full account details but getting details for a specific account (via [`GET /api/v1/accounts/:id`](https://github.com/tootsuite/documentation/blob/master/Using-the-API/API.md#accounts)) should be much lighter on the server.

lisp/mastodon-tl.el
```diff
@@ -512,13 +513,33 @@ LINK-TYPE is the type of link to produce."
      'keymap mastodon-tl--link-keymap
      'help-echo help-text)))
 
+(defun mastodon-tl--handle-to-account (handle)
```
hdurer: Name this something like "search account by handle"?  And place this in `mastodon-profile.el`?

lisp/mastodon-tl.el
```diff
@@ -512,13 +513,33 @@ LINK-TYPE is the type of link to produce."
...
+  "Return an account based on a users HANDLE.
+
+If the handle does not match a search return then retun NIL."
+  (let* ((handle (if (string= "@" (substring handle 0 1))
+                     (substring handle 1 (length handle))
+                   handle))
+         (matching-account
+          (remove-if-not
+           (lambda(x) (string= (cdr (assoc 'acct x)) handle))
```
hdurer: Do we need to worry about case sensitivity here?

lisp/mastodon-tl.el
```diff
@@ -513,13 +514,33 @@ LINK-TYPE is the type of link to produce."
...
 (defun mastodon-tl--do-link-action-at-point (position)
   (interactive "d")
   (let ((link-type (get-text-property position 'mastodon-tab-stop)))
     (cond ((eq link-type 'content-warning)
            (mastodon-tl--toggle-spoiler-text position))
           ((eq link-type 'hashtag)
            (mastodon-tl--show-tag-timeline (get-text-property position 'mastodon-tag)))
+          ((eq link-type 'handle)
+           ;; replace message with mastodon-profile--make-author-buffer
+           (message "its a handle: %s"
```
hdurer: In my cursory testing this seems to work.
Get rid of the `message` and instead actually show the profile.

test/mastodon-tl-tests.el
```diff
@@ -266,7 +266,7 @@ a string or a numeric."
  | Account 42 (@acct42@example.space) 2999-99-99 00:11:22
   ------------"))
 	(should (eq (get-text-property handle-location 'mastodon-tab-stop byline)
-                'shr-url))
+                'handle))
```
hdurer: We should probably also assert that there is the expected property with the account data (handle name at the moment) at the point.

lisp/mastodon-tl.el
alexjgriffith: Good call. Since the account is in the toot, we can check if the handle has 'toot-json before doing a search and extract the account from there.

lisp/mastodon-tl.el
alexjgriffith: That would work for byline handles, but do inline handles have id's in their meta info? For the time being we will look up the account id in 'toot-json, if it doesn't exist we can do a search.

alexjgriffith: I see, the ids are in the mentions, will save a lot of search calls.

hdurer (approved):

PR #191, opened 2018-03-17
Give `mastodon-media--inline-images` a saner interface.

hdurer:
Instead of making it search the whole buffer every time to find images to load, give it a range where this work should be done.
We then call this immediately after inserting a single status, notification, ...

There should be no big noticible difference - images might load a tiny bit sooner although I doubt you can see that.
This should be more efficient on large buffers although Alex didn't notice any problems when testing streamed buffers.
We should still do it as it make things easier to understand.  I was always worried about these global operations.


alexjgriffith (approved):
LGTM

Good call on setting the regions explicitly!

PR #193, opened 2018-03-26
More profile work

hdurer:
Add an alternative approach to user profile opening.

This way asks the user in the minibuffer for the handle and offering completion for all user handles in the current status but allowing the user to also enter any other handle to browse whichever account they wish.

This also
 - Cleans up some compiler warnings about profile code.
 - Creates a new minor mode for mastodon profile pages.  There we override the 'f' and 'F' keys to show following and followers respectively.
 - Modifies the followers/following pages to look very similar to the regular profile page (with the account header).

### DIFF
--- a/README.org
+++ b/README.org
@@ -77,37 +77,38 @@ Opens a =*mastodon-home*= buffer in the major mode so you can see toots. You wil
 
 **** Keybindings
 
-|--------------------------+-----------------------------------------------------------------------------------|
-| Key                      | Action                                                                            |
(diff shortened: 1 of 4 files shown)


alexjgriffith:
Works great! Could we use another key for viewing who a user is following? I would like to be able to favourite toots from the profile view.

alexjgriffith (approved):
Code LGTM.  Merging the toots following and followers really cleaned up the module.

PR #194, opened 2018-03-29
Fixes to boosting/favoriting.

hdurer:
I had issues that occasionally I couldn't unfavorite / unboost a toot.
Once that was fixed it annoyed me that the cursor would often jump on those actions (jumping to the byline with no need).


alexjgriffith (approved):
LGTM Tested it out in emacs 25.3 and everything seams good.

I like use of save-excursion.

PR #195, opened 2018-03-29
Tiny fixes to `mastodon-inspect--dump-json-in-buffer`

hdurer:
- Erase buffer before adding new contents
- Don't globally set `print-level` and `print-length`; just temporarily let them to the locally desired values.


alexjgriffith (approved):
LGTM

PR #197, opened 2018-04-03
Make most GET calls in the api async.

hdurer:
This stops mastodon from blocking all of emacs when you have a slow connection.

Since the display of a longer timeline could also be quite slow we now to the insertions in reverse order (i.e. starting from oldest) and redisplay after each toot. That gives the user some nice visual feedback as the toots are flowing into and down the buffer one by one.

I failed to update the mocking in a small number of tests which therefore had to be removed. I'll need to learn more about the mocking and stubbing in ert-tests to see if we can get this working if part of the arguments we want to match are very dynamic things like buffers.

### DIFF
--- a/lisp/mastodon-http.el
+++ b/lisp/mastodon-http.el
@@ -81,10 +81,10 @@ Authorization header is included by default unless UNAUTHENTICED-P is non-nil."
                       args
                       "&")))
         (url-request-extra-headers
-	 (append
-	  (unless unauthenticed-p
-	    `(("Authorization" . ,(concat "Bearer " (mastodon-auth--access-token)))))
-	  headers)))
+         (append
+          (unless unauthenticed-p
+            `(("Authorization" . ,(concat "Bearer " (mastodon-auth--access-token)))))
+          headers)))
     (with-temp-buffer
       (url-retrieve-synchronously url))))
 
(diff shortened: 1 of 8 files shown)


alexjgriffith:
It works, and the code looks clean apart from some function descriptions (which were also missing in the develop branch). However, the behaviour seams to be a little odd.

1. The response time seams to be noticeably longer than using `url-retreive-syncronouslty`
2. Rather than waiting for the request processing to be complete and then inserting all toots as a batch, the toots are inserted one at a time (this may be a preference thing). While individual toots are loading it is nearly impossible to navigate the buffer. This is an issue when requesting more toots.
3. When loading older toots they appear at the top of the buffer rather than the bottom. 
4. When requesting more toots its not clear that the call has been made. I accidentally made a request 4-5 times before I started seeing results, and the resulting repeated loads made the buffer unnavigable. Perhaps a line could be added at the top or bottom of the buffer indicating that the call has been made, or, to prevent multiple requests, some state could be added as buffer variable that flags when a request is in progress.

These comments are just a summary of my experience using the async buffers PR, and some may simply be based on preference. I will take a deeper dive into the PR to see if I can suggest where some of these changes may be implemented.

hdurer:
Yes, there were indeed quite a few issues.
In the end I basically gave up.  So much other stuff gets merged it's hard to keep this up-to-date.

I'll close this PR and see if I come back to this as some point.  The branch should remain available over in my repo.

PR #203, opened 2018-08-17
Use the email address of the current user as default

juergenhoetzel:


hdurer:
Nice!  

But there are some tests now failing.  We mock everything and if you add an argument, it doesn't match any more.   Yeah, brittle...

juergenhoetzel:
H Durer writes:

> Nice!
>
> But there are some tests now failing. We mock everything and if you add an argument, it doesn't match any more. Yeah, brittle...

I see: I changed the mocking accordingly.

hdurer (approved):
Great.  Thanks!

PR #198, opened 2018-04-04
Improvements on posting toots.

hdurer:
- Visual feedback in the buffer of:
  - Number of characters
  - Selected visibility
  - Content warning flag
- Can't post an empty toot.
- Changing visibility to something other than "public".

Still missing:
 - Attaching media
 - Enabling the NSFW toggling

PR #204, opened 2018-08-18
Add ability to set toot visibility

jdormit:
This adds the ability to change the visibility of toots. The default visibility is `public`, but that can be changed by setting the variable `mastodon-toot--default-visibility`.

### DIFF
--- a/lisp/mastodon-toot.el
+++ b/lisp/mastodon-toot.el
@@ -32,9 +32,16 @@
 (defvar mastodon-instance-url)
 (defvar mastodon-toot--content-warning nil)
 (defvar mastodon-toot--visibility)
-(defvar mastodon-toot--default-visibility "public"
+(defcustom mastodon-toot--default-visibility "public"
   "Default visiblity for toots.
-Valid values: \"public\", \"unlisted\", \"private\", \"direct\".")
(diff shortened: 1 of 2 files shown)


lisp/mastodon-toot.el
```diff
@@ -31,6 +31,11 @@
 
 (defvar mastodon-instance-url)
 (defvar mastodon-toot--content-warning nil)
+(defvar mastodon-toot--visibility)
+(defvar mastodon-toot--default-visibility "public"
```
hdurer: Make this a custimizable value?  (And then document it in the README file...)

hdurer (commented):
I like that the default visibility is changeable.
The rest is in conflict with my own PR #198 so I personally am obviously biased here...

jdormit:
Ah yup, your PR does what this does and more. Thanks for taking a look here! I will take a look at yours as well when I have time - I have a feeling I'll end up closing this PR in favor of yours since it is more extensive.

jdormit:
Alternatively, we could merge this one (since it is very similar to the approach that you took) and then merge in your changes afterward.

jdenen:
Pulled in with #210. Thanks for the PR @jdormit!

PR #213, opened 2019-02-15
OAuth2 support

codingquark:
Adds support for 2Factor Authentication (OAuth2).

New dependencies:
[oauth2](https://elpa.gnu.org/packages/oauth2.html)

Includes readme instructions on how to get relevant keys from mastodon instance and setup local variables.

Makes OAuth2 as default authentication mechanism. This is to be discussed and debated, I suppose.

I would also appreciate feedback on improving or changing test cases.


README.org
```diff
@@ -69,6 +65,40 @@ There is an option to have your user credentials (email address and password) sa
...
+*** 2 Factor Auth
+**** Overview
+After setting up =mastodon-instance-url= set the following in your
+=.emacs= or =customize=:
+
+1. =mastodon-auth-mechanism=
+2. =mastodon-oauth2-client-id=
+3. =mastodon-oauth2-client-secret=
+
+**** Do the following to obtain OAuth2 credentials:
```
hdurer: Isn't this exactly what `mastodon-client--register` does?  Why does the user have to do that manually now?

lisp/mastodon-auth.el
```diff
@@ -60,10 +81,11 @@ if you are happy with unencryped storage use e.g. \"~/authinfo\"."
...
 (defun mastodon-auth--generate-token ()
   "Make POST to generate auth token."
-  (if (or (null mastodon-auth-source-file)
-	  (string= "" mastodon-auth-source-file))
+  (if (and (or (null mastodon-auth-source-file)
+               (string= "" mastodon-auth-source-file))
+           (string= "plain" mastodon-auth-mechanism))
       (mastodon-auth--generate-token-no-storing-credentials)
-    (mastodon-auth--generate-token-and-store)))
+    (mastodon-oauth2--generate-token-and-store)))
```
hdurer: Does that mean we can no longer store the credentials when using `plain` mode?
Does that mean it goes through the token generation for each API call when using `plain`?

hdurer (commented):
I had just a quick first look.  Looks generally ok.  I don't understand if this causes functionality regressions for plain mode.

README.org
codingquark: I see, missed that one. I shall refactor.

lisp/mastodon-auth.el
codingquark: The OAuth2 library stores the credentials in its own file, which means it doesn't affect the way we store the credentials in `plain` mode. The method will check if the auth mechanism is `plain` or `oauth2`, and will call token generation procedure accordingly.

The OAuth2 library uses the existing tokens, refreshes them if needed, or generates new ones if not available. Hence, the calls will go through the procedure, but HTTP APIs will only be called when necessary.

codingquark:
> I had just a quick first look. Looks generally ok. I don't understand if this causes functionality regressions for plain mode.

I have run existing tests in my local machine, and they are passing. We probably will need more test coverage to increase confidence level. I have also done some manual testing on my own account (which of course should not be trusted, because I'm biased!)

codingquark:
Updated to use `mastodon-client` while generating oauth2 tokens. Please read commit message for details.

codingquark:
Travis is getting `Lisp nesting exceeds ‘max-lisp-eval-depth’` - what should I do about this?

Another error `at 149:1: error: "mastodon-oauth2--access-token" doesn't start with package's prefix "mastodon-auth"`. I have no idea how this works, but I am assuming I need to prefix it with `mastodon-auth`, will it be good enough?

codingquark:
```
In ‘lisp/mastodon-auth.el’:
  at 7:39: error: Package oauth2 is not installable.
```

Is it because oauth2.el is at 0.11 and not considered stable? Because it is surely available from `(source gnu)` : https://elpa.gnu.org/packages/oauth2.html

README.org
```diff
@@ -69,6 +65,22 @@ There is an option to have your user credentials (email address and password) sa
...
+1. =mastodon-auth-mechanism=
+
+**** Set up
+#+BEGIN_SRC emacs-lisp
+  (setq mastodon-auth-mechanism "oauth2")
+#+END_SRC
+
+The OAuth2 library uses =plstore= to securely store the credentials.
+However, =mastodon-oauth2-client-secret= needs to be stored in a safe
+place - encryption is recommended.
```
hdurer: That isn't very actionable advice.  Where is that stored and how does the user ensure it is encrypted?

README.org
```diff
@@ -69,6 +65,22 @@ There is an option to have your user credentials (email address and password) sa
...
+=.emacs= or =customize=:
+
+1. =mastodon-auth-mechanism=
+
+**** Set up
+#+BEGIN_SRC emacs-lisp
+  (setq mastodon-auth-mechanism "oauth2")
+#+END_SRC
+
+The OAuth2 library uses =plstore= to securely store the credentials.
```
hdurer: Can you add a link somewhere appropriate to the oauth2 library (plus maybe a note which version)?

E.g. I don't use melpa, so I'll need to find the library and manually add it to me config files repo.

hdurer: Is that plstore location configurable?  Maybe mention the appropriate var?

lisp/mastodon-auth.el
```diff
@@ -108,6 +112,17 @@ Reads and/or stores secres in `MASTODON-AUTH-SOURCE-FILE'."
       (when (functionp (plist-get credentials-plist :save-function))
         (funcall (plist-get credentials-plist :save-function))))))
 
+;; OAuth2
+(defun mastodon-auth-oauth2--generate-token-and-store ()
+  "Generate OAuth2 token.
+
+Reads and/or stores secres in `OAUTH2-TOKEN-FILE'."
```
hdurer: s/secres/secrets/

lisp/mastodon-http.el
```diff
@@ -83,10 +84,17 @@ Authorization header is included by default unless UNAUTHENTICED-P is non-nil."
         (url-request-extra-headers
 	 (append
 	  (unless unauthenticed-p
-	    `(("Authorization" . ,(concat "Bearer " (mastodon-auth--access-token)))))
+	    `(("Authorization" . ,(concat "Bearer " (if (string= "oauth2" mastodon-auth-mechanism)
```
hdurer: Is this header needed at all for oauth2 or should the `(string= "oauth2" mastodon-auth-mechanism)` check go into the `unless` above to totally bypass this in the new flow?

hdurer:
I now tried this as well and it works for me (yay!).

One issue I notice is that with the old code I can switch between servers by just changing `mastodon-instance-url` and doing `M-x mastodon` again.  With the oauth2 code this does not seem to use the correct client secret / generate a new client; instead the token generation page basically told me that things failed due to an unknown client being used.

hdurer:
BTW: Travis seems to be failing and there is a new warning
```
In ‘lisp/mastodon-auth.el’:
  at 7:39: error: Package oauth2 is not installable.
```

hdurer:
> One issue I notice is that with the old code I can switch between servers by just changing `mastodon-instance-url` and doing `M-x mastodon` again. With the oauth2 code this does not seem to use the correct client secret / generate a new client; instead the token generation page basically told me that things failed due to an unknown client being used.

Looking at the code, I have no explanation for this happening.  It should keep the secrets separate by instance url...

So to try again I tried this with a third instance and that fails differently - with an infinite loop like this:

```
  mastodon-http--post("https://toot.love/api/v1/apps" (("client_name" . "mastodon.el") ("redirect_uris" . "urn:ietf:wg:oauth:2.0:oob") ("scopes" . "read write follow") ("website" . "https://github.com/jdenen/mastodon.el")) nil :unauthenticated)
  mastodon-client--register()
  (set-buffer (mastodon-client--register))
  (save-current-buffer (set-buffer (mastodon-client--register)) (goto-char (point-min)) (re-search-forward "^$" nil 'move) (let ((
  mastodon-client--fetch()
  (let ((plstore (
  mastodon-client--store()
  (or (mastodon-client--read) (mastodon-client--store))
  (setq client-details (or (mastodon-client--read) (mastodon-client--store)))
  (if client-details nil (setq client-details (or (
  (let ((client-details (
  mastodon-client()
  (plist-get (mastodon-client) :client_id)
  (oauth2-auth-and-store (concat mastodon-instance-url "/oauth/authorize") (concat mastodon-instance-url "/oauth/token") "read write follow" (plist-get (mastodon-client) :client_id) (plist-get (mastodon-client) :client_secret))
  mastodon-auth-oauth2--generate-token-and-store()
  mastodon-auth-oauth2--access-token()
  (oauth2-url-retrieve-synchronously (mastodon-auth-oauth2--access-token) url url-request-method url-request-data)
  (if (string= "oauth2" mastodon-auth-mechanism) (oauth2-url-retrieve-synchronously (mastodon-auth-oauth2--access-token) url url-request-method url-request-data) (url-retrieve-synchronously url))
  (progn (if (string= "oauth2" mastodon-auth-mechanism) (oauth2-url-retrieve-synchronously (
  (unwind-protect (progn (if (
  (save-current-buffer (set-buffer temp-buffer) (unwind-protect (progn (
  (let ((temp-buffer (
  (let ((url-request-method "POST") (url-request-data (
  mastodon-http--post("https://toot.love/api/v1/apps" (("client_name" . "mastodon.el") ("redirect_uris" . "urn:ietf:wg:oauth:2.0:oob") ("scopes" . "read write follow") ("website" . "https://github.com/jdenen/mastodon.el")) nil :unauthenticated)
  mastodon-client--register()
```

README.org
codingquark: oauth2.el is available on elpa. Do you think we should link it?

README.org
codingquark: Updated, how is it now?

README.org
codingquark: Added info on steps.

lisp/mastodon-http.el
codingquark: I have updated post and get both.

hdurer:
Woot.  I got it to work for me now.

These are the changes I made locally to not get the infinite loop.
Basically, trying to generate the token it tried to do a post which would require a token.  So, now for just the token generation I bind `mastodon-auth-mechanism` to `'plain` to avoid that.  I guess alternatively, you should check for unauthenticated also when making the check whether to use the normal `url-retrieve-synchronously` or the oauth2 variant.  That's your call.  (But *do* note that you should use `or` in the `unless` to correctly avoid setting the Authorization header.

```
diff --git a/lisp/mastodon-client.el b/lisp/mastodon-client.el
index 52eaae9..9463e82 100644
--- a/lisp/mastodon-client.el
+++ b/lisp/mastodon-client.el
@@ -33,6 +33,7 @@
 (defvar mastodon-instance-url)
 (autoload 'mastodon-http--api "mastodon-http")
 (autoload 'mastodon-http--post "mastodon-http")
+(autoload 'mastodon-auth-mechanism "mastodon-auth")
 
 
 (defcustom mastodon-client--token-file (concat user-emacs-directory "mastodon.plstore")
@@ -45,14 +46,15 @@
 
 (defun mastodon-client--register ()
   "POST client to Mastodon."
-  (mastodon-http--post
-   (mastodon-http--api "apps")
-   '(("client_name" . "mastodon.el")
-     ("redirect_uris" . "urn:ietf:wg:oauth:2.0:oob")
-     ("scopes" . "read write follow")
-     ("website" . "https://github.com/jdenen/mastodon.el"))
-   nil
-   :unauthenticated))
+  (let ((mastodon-auth-mechanism 'plain))
+    (mastodon-http--post
+     (mastodon-http--api "apps")
+     '(("client_name" . "mastodon.el")
+       ("redirect_uris" . "urn:ietf:wg:oauth:2.0:oob")
+       ("scopes" . "read write follow")
+       ("website" . "https://github.com/jdenen/mastodon.el"))
+     nil
+     :unauthenticated)))
 
 (defun mastodon-client--fetch ()
   "Return JSON from `mastodon-client--register' call."
diff --git a/lisp/mastodon-http.el b/lisp/mastodon-http.el
index 345f7f4..ddade6d 100644
--- a/lisp/mastodon-http.el
+++ b/lisp/mastodon-http.el
@@ -83,8 +83,8 @@ Authorization header is included by default unless UNAUTHENTICED-P is non-nil."
                       "&")))
         (url-request-extra-headers
 	 (append
-	  (unless (and unauthenticed-p
-                       (string= "oauth2" mastodon-auth-mechanism))
+          (unless (or unauthenticed-p
+                      (string= "oauth2" mastodon-auth-mechanism))
 	    `(("Authorization" . ,(concat "Bearer " (mastodon-auth--access-token)))))
 	  headers)))
     (with-temp-buffer
```

README.org
hdurer: I found and use https://github.com/emacsmirror/oauth2.git in my own setup.

hdurer: Seems the var is `oauth2-token-file`, we could/should mention that explicitly here.

Actually, could we just temporarily around oauth operations let `oauth2-token-file` to be the same value as `mastodon-auth-source-file` and thereby use only one secrets storage?

README.org
```diff
@@ -69,6 +65,24 @@ There is an option to have your user credentials (email address and password) sa
...
+*** 2 Factor Auth
+**** Overview
+After setting up =mastodon-instance-url= set the following in your
+=.emacs= or =customize=:
+
+1. =mastodon-auth-mechanism=
+
+**** Set up
+#+BEGIN_SRC emacs-lisp
+  (setq mastodon-auth-mechanism "oauth2")
```
hdurer: Actually it seem that  `mastodon-auth-mechanism` should be a symbol (that's what you define as options in the defcustom.  So this should be `'oauth2` and not `"oauth2"`.

It would be clearer, if you had consistently use `(eq 'oauth2 mastodon-auth-mechanism)` instead of doing a `(string= ...`.

lisp/mastodon-http.el
```diff
@@ -82,21 +83,31 @@ Authorization header is included by default unless UNAUTHENTICED-P is non-nil."
                       "&")))
         (url-request-extra-headers
 	 (append
-	  (unless unauthenticed-p
+	  (unless (and unauthenticed-p
```
hdurer: As mentioned in the other big comment, I think this should be `or` not `and`.

lisp/mastodon-http.el
```diff
@@ -82,21 +83,31 @@ Authorization header is included by default unless UNAUTHENTICED-P is non-nil."
...
+                       (string= "oauth2" mastodon-auth-mechanism))
 	    `(("Authorization" . ,(concat "Bearer " (mastodon-auth--access-token)))))
 	  headers)))
     (with-temp-buffer
-      (url-retrieve-synchronously url))))
+      (if (string= "oauth2" mastodon-auth-mechanism)
```
hdurer: As mentioned in other comment: Don't use `oauth2-url-retrieve-synchronously` when making an unauthenticated call.

hdurer (commented):
I really like this... only a few more comments.

README.org
codingquark: I had used `defvaralias` for a while. But I decided to go against it. Since this is a generic library which can be used as a dependency by other packages, I think aliasing it will not be a good idea.

Is there any alternative approach so that the global env is not spoiled?

lisp/mastodon-http.el
codingquark: Instead of this, I am going with the modification in `mastodon-client--register`. This will make sure the settings are changed only when needed and will keep logical clarity intact.

In my point of view, we don't want to use oauth2 to register the client. Which has nothing to do with generic post requests directly. Although it is true that not making calls to `oauth2-url-retrieve-synchronously` is a solution, it is at a place which is too broad.

I hope I am making the reasoning clear.

codingquark: Now that I'm thinking about it, fixing it at the POST call will prevent future bugs too.

README.org
hdurer: Nit: `'oauth` - no more strings.

lisp/mastodon-auth.el
```diff
@@ -60,10 +61,13 @@ if you are happy with unencryped storage use e.g. \"~/authinfo\"."
 
 (defun mastodon-auth--generate-token ()
   "Make POST to generate auth token."
-  (if (or (null mastodon-auth-source-file)
-	  (string= "" mastodon-auth-source-file))
-      (mastodon-auth--generate-token-no-storing-credentials)
-    (mastodon-auth--generate-token-and-store)))
+  (cond ((and (or (null mastodon-auth-source-file)
+                  (string= "" mastodon-auth-source-file))
+              (string= "plain" mastodon-auth-mechanism))
```
hdurer: Nit: `(eq 'plain mastodon-auth-mechanism)`

test/mastodon-auth-tests.el
```diff
@@ -4,7 +4,8 @@
   "Should make `mastdon-http--post' request to generate auth token."
   (with-mock
    (let ((mastodon-auth-source-file "")
-	 (mastodon-instance-url "https://instance.url"))
+	 (mastodon-instance-url "https://instance.url")
+         (mastodon-auth-mechanism "plain"))
```
hdurer, line 8: Nit: `'plain`   (and a few more elsewhere...)

test/mastodon-oauth2-tests.el
```diff
@@ -0,0 +1,49 @@
+(require 'el-mock)
+
+(ert-deftest oauth2-is-called-while-generating-token ()
+  "Should call `mastodon-auth-oauth2--generate-token-and-store' to generate auth token."
+  :expected-result :failed
+  (with-mock
+   (let ((mastodon-instance-url "https://instance.url")
+         (mastodon-auth-mechanism "oauth2"))
```
hdurer: Nit: `'oauth2`  (and 2 more below)

lisp/mastodon-auth.el
hdurer: Ping!  This still says "secres"

lisp/mastodon-client.el
```diff
@@ -45,14 +46,15 @@
...
   "POST client to Mastodon."
-  (mastodon-http--post
-   (mastodon-http--api "apps")
-   '(("client_name" . "mastodon.el")
-     ("redirect_uris" . "urn:ietf:wg:oauth:2.0:oob")
-     ("scopes" . "read write follow")
-     ("website" . "https://github.com/jdenen/mastodon.el"))
-   nil
-   :unauthenticated))
+  (let ((mastodon-auth-mechanism 'plain))
```
hdurer, line 52: Is this now actually still needed?  You also applied the other fix (checking whether the call is unauthenticated), right?

lisp/mastodon-http.el
```diff
@@ -82,21 +83,32 @@ Authorization header is included by default unless UNAUTHENTICED-P is non-nil."
                       "&")))
         (url-request-extra-headers
 	 (append
-	  (unless unauthenticed-p
+	  (unless (or unauthenticed-p
+                      (eq 'oauth2 mastodon-auth-mechanism))
```
hdurer, line 88: That indentation looks weird.

When you are done, can you run `M-x untabbify` on all touched files to get rid of these pesky tab characters?

hdurer (commented):
Looking good.
Just some nits left as comments.

I still want to use this code longer than 30s but in general I am really happy about this.
We should come up with some path to get rid of the old auth mechanism - or is there any use for it?

hdurer:
Oh, I see that there are now compiler warnings in `mastodon-auth.el`.  At least once these identified actual bugs so could we fix those?

hdurer:
About the Travis failures: Do we need to add the dependency on oauth2 to the `Cask` file?  https://cask.readthedocs.io/en/latest/guide/dsl.html#dependencies

lisp/mastodon-http.el
codingquark: @hdurer this is regarding the fix being put at two places.

hdurer:
So I used the latest code for a while and found no issues.

If only we could get Travis to be happy as well...

hdurer:
I can locally get the package lint step to not error, when I remove the `(oauth2 "0.11")` from `/mastodon-auth.el`

codingquark:
Indeed, and I am blaming my ignorance of Cask for the CI not being happy.

hdurer:
> Indeed, and I am blaming my ignorance of Cask for the CI not being happy.

If you remove that item we get rid of the error.  There is still the warning which makes Travis also barf, but I just raised that as a separate issue.  It's not related to this code.

codingquark:
If I remove the line from `mastodon-auth.el`, how will the dependencies work?

hdurer:
> If I remove the line from `mastodon-auth.el`, how will the dependencies work?

Just to clarify: I mean that you remove it from the `Package-Requires:` comment in the header.  
I assume that that is meant for the package dependency but the package linter is obviously unhappy with it, so removing it seems to be a sensible way out.

hdurer:
OK, so Travis is happy except for the preexisting issue (see issue #219).

Now we are blocked on PR #220...  @jdenen is assigned to that one.

hdurer (approved):
As mentioned before, I am happy with this code. nwo.

codingquark:
Cleaned up those nasty warnings too.

jdenen:
Fixed all the merge conflicts that I created. Also, #219 has been fixed thanks to #220, so CI is passing again.

I'll be taking a closer look this in the next couple of days. Thanks for the PR!

arpunk:
Any news on this?

jdenen:
Sorry for my lagging attention. Thanks @codingquark @hdurer

codingquark:
> Sorry for my lagging attention. Thanks @codingquark @hdurer

That's cool! Finally merged!

Issue #211, opened 2019-02-04
Newbie problems ... 


fenix011:
I'm getting some error messages when installing the package via ELPA. The shown messages are :

```
Leaving directory `/home/jordi/.emacs.d/elpa/mastodon-20180811.220'

Compiling file /home/jordi/.emacs.d/elpa/mastodon-20180811.220/mastodon-auth.el at Mon Feb  4 12:46:09 2019
Entering directory `/home/jordi/.emacs.d/elpa/mastodon-20180811.220/'
mastodon-auth.el:111:1:Warning: Unused lexical variable `json-array-type'
mastodon-auth.el:111:1:Warning: Unused lexical variable `json-key-type'
mastodon-auth.el:111:1:Warning: Unused lexical variable `json-object-type'

In end of data:
mastodon-auth.el:151:1:Warning: the function `json-read-from-string' is not
    known to be defined.

Compiling file /home/jordi/.emacs.d/elpa/mastodon-20180811.220/mastodon-autoloads.el at Mon Feb  4 12:46:10 2019

Compiling file /home/jordi/.emacs.d/elpa/mastodon-20180811.220/mastodon-client.el at Mon Feb  4 12:46:10 2019
mastodon-client.el:57:1:Warning: Unused lexical variable `json-array-type'
mastodon-client.el:57:1:Warning: Unused lexical variable `json-key-type'
mastodon-client.el:57:1:Warning: Unused lexical variable `json-object-type'

In end of data:
mastodon-client.el:112:1:Warning: the function `json-read-from-string' is not
    known to be defined.

Compiling file /home/jordi/.emacs.d/elpa/mastodon-20180811.220/mastodon-http.el at Mon Feb  4 12:46:10 2019

Compiling file /home/jordi/.emacs.d/elpa/mastodon-20180811.220/mastodon-inspect.el at Mon Feb  4 12:46:10 2019

Compiling file /home/jordi/.emacs.d/elpa/mastodon-20180811.220/mastodon-media.el at Mon Feb  4 12:46:10 2019

Compiling file /home/jordi/.emacs.d/elpa/mastodon-20180811.220/mastodon-notifications.el at Mon Feb  4 12:46:10 2019

Compiling file /home/jordi/.emacs.d/elpa/mastodon-20180811.220/mastodon-pkg.el at Mon Feb  4 12:46:10 2019

[...]

PR #214, opened 2019-02-15
Explicitly require json.el in files where we dynamically bind its vars.

hdurer:
As pointed out in issue #211 the compiler actually warned the user about `json-array-type` etc. being unused lexical vars.

Let's hope that this fixes the errors reported with reading json.

### DIFF
--- a/lisp/mastodon-auth.el
+++ b/lisp/mastodon-auth.el
@@ -31,6 +31,7 @@
 
 (require 'plstore)
 (require 'auth-source)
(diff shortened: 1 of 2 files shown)


hdurer:
I don't understand the failure in travis.  🤷

hdurer:
> I don't understand the failure in travis. 🤷

I'm pretty sure that's issue #219 and is not my fault.

jdenen (approved):

PR #231, opened 2020-05-03
Rewrite `mastodon-auth--access-token` so it handles errors.

ieure:
Currently, `mastodon-auth--access-token` unconditionally returns the
value of the `:access_token` key from the response of
`(mastodon-auth--get-token)`.  This causes problems when there was an
error getting the token, for example, if you enter the wrong
password.  If a token couldn’t be retrieved, the JSON looks like:

    (:error "invalid_grant"
     :error_description "The provided authorization grant is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client.")

Since there is no `:access_token` key, `mastodon-auth--access-token`
returns `nil`, which results in a broken header in the next request:

    Authorization: Bearer

Which causes the whole thing to freeze Emacs until you mash `C-g`.

This commit rewrites the function to handle that case; to explicitly
signal an error for *any* response that isn’t expected; to use
`if-let`, which allows the temporary `token` variable to be
eliminated; uses `pcase` to determine what kind of response was
received; and adds ERT tests for all these cases.

### DIFF
--- a/lisp/mastodon-auth--test.el
+++ b/lisp/mastodon-auth--test.el
@@ -1,0 +1,47 @@
+;;; mastodon-auth--test.el --- Tests for mastodon-auth  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2020  Ian Eure
+
+;; Author: Ian Eure <ian@retrospec.tv>
+;; Version: 0.9.0
+;; Homepage: https://github.com/jdenen/mastodon.el
+;; Package-Requires: ((emacs "24.4"))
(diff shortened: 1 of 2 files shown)


martianhi:
@jdenen is there any chance of merging this? The version currently on MELPA is broken for many users, and @ieure's changes fix it.

if not, are you willing to pass your repo over to me or someone else, or consent to let MELPA take from my fork, which incorporates this fix and numerous other things?

